PR #32, opened 2019-01-03
Add a /ping endpoint

Peltoche:
Hi!

I need to create a kubernetes pod with gotenberg and there is currently no easy way to set the liveness and readiness probs. In order to solve that issue I propose this ping endpoint which just return a 200 once the API is available.

It can be used to check the service liveness un docker-compose and is very useful for kubernetes (liveness and readiness probs).

### DIFF
--- a/internal/app/api/api.go
+++ b/internal/app/api/api.go
@@ -67,6 +67,7 @@ func setup() *echo.Echo {
 			return nil
 		}
 	})
+	e.GET("/ping", func(c echo.Context) error { return nil })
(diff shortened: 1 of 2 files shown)


gulien:
Indeed, I've been using Helm chart lately and this is a must-have. Thanks 👍

gulien:
I'll update the documentation and create a new release 🎉

Peltoche:
So it will be the version "3.1.3" for the docker image?

gulien:
3.2.0 (new "feature")

PR #45, opened 2019-01-31
Fix for #44 promise wait for fonts

### LINKED ISSUE #44: [BUG] HTML-to-PDF sometimes lose some parts of HTML
## Expected Behavior

PDF generated from HTML contain all elements of HTML.

## Current Behavior

Sometimes (1 of 10 in average) parts of HTML elements lost in resulted PDF. Biggest problem is that response status code is 200.

## Steps to Reproduce (for bugs)

1. Create `index.html` with content https://gist.github.com/server-may-cry/57b062186f648a16f6e4c5c7ac354e38
2. `curl --request POST --url http://gotenberg:3000/convert/html --header 'Content-Type: multipart/form-data' --form files=@index.html -o result.pdf`
3. See file size
4. Repeat 10-20 times each time check file size

## Context

No difference in logs from container for failed and ok requests
```
ok
gotenberg_1                | {"time":"2019-01-28T17:56:49.563200149Z","id":"","remote_ip":"172.18.0.16","host":"gotenberg:3000","method":"POST","uri":"/convert/html","user_agent":"curl/7.52.1","status":200,"error":"","latency":423976853,"latency_human":"423.976853ms","bytes_in":8501,"bytes_out":28024}

failed
gotenberg_1                | {"time":"2019-01-28T18:02:20.799825196Z","id":"","remote_ip":"172.18.0.16","host":"gotenberg:3000","method":"POST","uri":"/convert/html","user_agent":"curl/7.52.1","status":200,"error":"","latency":566488490,"latency_human":"566.48849ms","bytes_in":8501,"bytes_out":28024}
```

## Your Environment

* Version used: tested on 3.2 and 4 (on version 4 it happens very rarely)
* Operating System and version: docker image
* Link to your project:

server-may-cry:
**Summary**

Error: `runtime.ExceptionDetails: Uncaught exception at 4:47: SyntaxError: Unexpected token %`

**Closing issues**

Fixes #44

**Checklist**

- [ ] Have you followed the guidelines in our [CONTRIBUTING](CONTRIBUTING.md) guide?
- [ ] Have you written new tests for your core changes, as applicable?
- [x] Have you updated the documentation (`make doc`)?
- [x] This change has comments for package types, values, functions, and non-obvious lines of code

### DIFF
--- a/internal/pkg/printer/html.go
+++ b/internal/pkg/printer/html.go
@@ -107,10 +107,13 @@ func (html *HTML) Print(destination string) error {
 		document.fonts.ready.then(function () {
 			resolve('fonts loaded');
 		});
-		setTimeout(resolve.bind(resolve, 'timeout'), %.0f);
+		setTimeout(resolve.bind(resolve, 'timeout'), 500);
 	});`
 	scriptArg := runtime.NewEvaluateArgs(script).SetAwaitPromise(true)
 	returnObj, _ := c.Runtime.Evaluate(html.Context, scriptArg)
+	if returnObj.ExceptionDetails != nil {
+		return fmt.Errorf("script evaluated with exception: %+v", returnObj.ExceptionDetails)
+	}
 	loadFontsResult := string(returnObj.Result.Value)
 	if strings.Contains(loadFontsResult, "timeout") {
 		return errors.New("timed out loading fonts")


gulien:
Great thank you! So I didn’t set a proper timeout lol 👍

The error shown here is because of the expression ´%.0f´, right?

500ms is an acceptable timeout in you case?

server-may-cry:
> The error shown here is because of the expression ´%.0f´, right?

Yes, here should be `numeric` (float or int) value. Probably you plan to use `fmt.Sprintf()`

> 500ms is an acceptable timeout in your case?

For me is important to receive 500 status code and error in log. To know reason of problem and be able to resolve it somehow.
I didn't test this tool on production yet, maybe this error even won't be reproduced because of fast internet connection.
Best way would be to let configure this value via `gotenberg --html-fonts-timeout` with default value.
Just tell me if you think that here should be different value. Maybe I was too optimistic with 500ms default value

gulien:
> For me is important to receive 500 status code and error in log. To know reason of problem and be able to resolve it somehow.

I totally agree!

I'll add a form value in order to change the font loading timeout.

Thank you for your time 😄

PR #68, opened 2019-04-22
Fix google chrome env var error message

hspak:
**Summary**

This PR fixes a copy and paste error with the error message. I saw an 
```
error: DEFAULT_WAIT_TIMEOUT: wrong value: want "0" or "1" got true
```
When I should have seen 
```
error: DISABLE_GOOGLE_CHROME: wrong value: want "0" or "1" got true
```
when running gotenberg with `DISABLE_GOOGLE_CHROME=true`


gulien:
My bad! Thank you :)

Issue #65, opened 2019-04-05
Unsure about security


aurelien-clu:
Hello,
Nice work and documentation.

## Expected situation imho

Checking the `Dockerfile` and this repository code should be enough to have a pretty good idea whether this is secured or infected. (e.g. `data leakage`)

## Current situation

In your `Dockerfile`:

```
FROM thecodingmachine/gotenberg:3.2.0 AS hack
```

I am guessing that to reduce your `Dockerfile` you reuse a previous build but... this also hides what your previous build is based on or what is inside it, does not it?

Unless one investigate this particular container and look for issues, which is obviously harder that checking if a `Dockerfile` contains (more obvious) security issues.

## Possible solution

Would it be possible to have a longer `Dockerfile` without any `hack`, some says `explicit is better than implicit` :wink: 

Otherwise I guess I could try to reuse your `Dockerfile` from `tag` 3.2.0 and add it at the top of your latest one.

I am open to hear your point of view, I may be missing something.

aurelien-clu:
Ok I saw your comments in the `Dockerfile` and in https://github.com/Unitech/pm2/issues/4127, I guess I need to verify this previous image :/ (no offense :stuck_out_tongue: )

gulien:
Hello @aurelien-clu 

FYI, the previous `Dockerfile` is actually very close to the current one 😄

PR #72, opened 2019-05-08
Install pm2 via npm

nick-jones:
As noted in the Dockerfile, the previous route to install pm2 was broken at
some point. As a result, pm2 related files are now being copied over from an
image created prior to the breakage.

This change now installs pm2 via npm, which is the recommended route for this.
A small adjustment had to be made in the processManager code; an
argument was being passed with a space in it. For reasons that I've not been
able to fully trace, this no longer works. It should be noted that passing
arguments in this way can result in undesirable behaviour - in this case I
believe "--interpreter none" was being passed as a single argument, rather than
"--interpreter", "none". I've adjusted this to use an equals - splitting into 2
separate strings works well too.

Refs #65

### DIFF
--- a/build/base/Dockerfile
+++ b/build/base/Dockerfile
@@ -1,4 +1,3 @@
-FROM thecodingmachine/gotenberg:3.2.0 AS hack
 FROM debian:9.5-slim
 
 # |--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


gulien:
Great 👍! Thank you very much for this PR.

I’ve struggled with this issue for a while now and I was not expecting someone to solve it 😁

PR #75, opened 2019-05-23
skip logging when healthcheck url is called

### LINKED ISSUE #74: Possible to turn off logging for `/ping` ..?
Context: we run gotenburg in a kubernetes cluster as part of an overall deployment which includes other microservices that sit "in front" of gotenburg. Currently gotenburg is running fine, but when we go to pull logs they are almost completely unusable because every time we've healthcheck-ed it via a `/ping` that adds a log line to the output. 

Obviously we can pull the logs and grep, but it is a bit annoying given that we have no need to understand that a ping happened cause we can pull that information from kubernetes events.

compleatang:
This PR closes #74.

I wasn't sure organizationally the best fit in terms of location for the skip function, so I added it into middleware.go with the other middleware functions. 

I also took the opportunity to add the `/ping` endpoint as a constant just to DRY it up a bit. Tested locally and it works as expected.

### DIFF
--- a/build/docs/content/03-environment-variables.md
+++ b/build/docs/content/03-environment-variables.md
@@ -31,3 +31,11 @@ It takes a string representation of a float as value (e.g `"2.5"` for 2.5 second
 
 > The default timeout may also be overridden per request thanks to the form field `waitTimeout`.
 > See the [timeout section](#timeout).
+
+## Disable logging on healthcheck
+
+By default, the API will add a log entry when the [healthcheck endpoint](#ping) is called.
+
+You may turn off this logging so as to avoid unnecessary entries in your logs with the environment variable `DISABLE_HEALTHCHECK_LOGGING`.
+
+This environment variable operates in the same manner as the `DISABLE_GOOGLE_CHROME` and `DISABLE_UNOCONV` variables operate in that it accepts the strings `"0"` or `"1"` as values. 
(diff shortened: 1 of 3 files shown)


gulien:
Hello @compleatang,

Thank you for this PR! I wonder if it should be enable/disable through an environment variable like `SKIP_HEALTHCHECK_LOGGING`

compleatang:
@gulien should be all set now.

gulien:
Actually the environment variable should be retrieved in the [main.go](https://github.com/thecodingmachine/gotenberg/blob/master/cmd/gotenberg/main.go). See how I do it with `DEFAULT_WAIT_TIMEOUT` to understand how I use it in the API afterward 😄

compleatang:
Pending tests passing should be all set now.

compleatang:
Not really sure what's happening with that test failure. I cannot see why the simple change would cause that test to fail and on my local machine I cannot recreate the failure. Unfortunately I do not have permissions to rerun it to see if there is a kink in the matrix.

compleatang:
FWIW, I pushed a commit that mirrors origin/master and the travis job failed at the same place -> https://travis-ci.org/compleatang/gotenberg/jobs/537723905

gulien:
Sorry for the delay @compleatang. I'll try to take a look at your issue next week, I'm quite busy atm.

compleatang:
No prob!

Issue #87, opened 2019-06-12
Disk space leak


JWillow:
Hello,

I remark an issue about the space disk. When a timeout occurs, the process is stopped but the workdir inside the container isn't clear.

gulien:
By timeout, do you mean the API response (504)? 

If so, it should be cleared as I've create a handler for that: https://github.com/thecodingmachine/gotenberg/blob/master/internal/app/api/middleware.go#L44

guanchaoli:
> I had the code response 500 with this message
> 
> `{"time":"2019-06-12T07:46:09.073828582Z","id":"","remote_ip":"172.17.0.1","host":"10.0.2.4:8091","method":"POST","uri":"/convert/office","user_agent":"Java/1.8.0_212","status":500,"error":"code=500, message=unoconv: signal: killed","latency":10119491393,"latency_human":"10.119491393s","bytes_in":3083918,"bytes_out":38}`
> 
> After that I discover the parameter _waitTimeout_, I give 20 (more than the default (10)) and the conversion process finished properly.
> 
> When I had the code 500, the document still stored inside the container.

Could you tell me how to add this waitTimeout parameter?

PR #88, opened 2019-06-13
Feature/run non root and log startup

tbflw:
**Summary**

This PR fixes/implements the following **bugs/features**

* [X] Run as dedicated non-root user
* [X] Updates base debian image to newest 9.x version to get rid of some vulnerabilities
* [X] Adds configurable debug logging of starting the different processes to make debugging easier

We are running gotenberg in our K8S cluster and we don't want any pods to run as root user. Specifying a non-root user made the container fail starting up. First step to debug this was to add some debug logging allowing the stdout and stderr of process startup to be logged. 

This showed that the error was in fact missing permissions for the user to create the /.pm2 directory. Therefore we have changed the base image to create a dedicated user `gotenberg` with user id 1001 and a directory `/gotenberg` owned by this user. This user is used in the image and the `PM2_HOME` env var is set in the image to be `/gotenberg/.pm2` where the user has permission to write.

Debug logging of output may be controlled via a new environment variable `DEBUG_PROCESS_STARTUP` 

Furthermore, the resulting docker image has a lot of vulnerabilities, some of which could be removed by updating to a newer base Debian image.

**Test plan (required)**

1. `make gotenberg` to start the container 
2. Run the curl example from the doc to check pdf generation from html works

**Checklist**

- [X] Have you followed the guidelines in our [CONTRIBUTING](CONTRIBUTING.md) guide?
- [X] Have you lint your code locally prior to submission (`make lint`)?
- [/] Have you written new tests for your core changes, as applicable?
- [X] Have you successfully ran tests with your changes locally (`make tests`)?
- [X] Have you updated the documentation (`make doc`)?
- [X] I have squashed any insignificant commits
- [X] This change has comments for package types, values, functions, and non-obvious lines of code


markeissler:
@tbflw Thanks for doing this work. Looking into supporting a non-root account for running gotenberg was on my list of things to do as well.

With regard to setting the runtime uid: I'd like to be able to set the user higher up in the image build process or possibly leave it as late as startup: specify uid:gid as build args or env vars. This would require updating permissions with an entrypoint script.

Support for logging throughout the run process would be helpful as well. I'm currently debugging a `500: "Internal Server Error"` and have no further information to understand why this has happened with a 13MB docx file.

gulien:
Thank you very much for this PR @tbflw!

@markeissler might be related with #87?

build/package/Dockerfile
```diff
@@ -19,7 +19,11 @@ ENV GOOS=linux \
 WORKDIR /gotenberg
 
 # Copy our source code.
-COPY . .
+COPY internal ./internal
+COPY cmd ./cmd
+COPY go.sum go.sum
+COPY go.mod go.mod
+
```
gulien: As this is an "intermediate" image, is this really necessary? Or it's just for speeding up the build process?

cmd/gotenberg/main.go
```diff
@@ -49,37 +50,31 @@ func mustParseEnvVar() *api.Options {
...
-			os.Exit(1)
-		}
-		opts.EnableUnoconvEndpoints = v != "1"
-	}
-	if v, ok := os.LookupEnv(disableHealthcheckLoggingEnvVar); ok {
-		if v != "1" && v != "0" {
-			notify.ErrPrint(fmt.Errorf("%s: wrong value: want \"0\" or \"1\" got %v", disableHealthcheckLoggingEnvVar, v))
-			os.Exit(1)
+	//checkBoolEnv is a convenience function for reading an env var with a bool value where `1` is true and `0` is false
+	checkBoolEnv := func(name string) bool {
```
gulien: 👍

internal/pkg/pm2/pm2.go
```diff
@@ -76,8 +83,43 @@ func (m *processManager) pm2(p Process, cmdName string) error {
...
+		if err != nil {
+			return fmt.Errorf("failed getting Chrome stderr: %v", err)
+		}
+		chromeStdOut, err := cmd.StdoutPipe()
+		if err != nil {
+			return fmt.Errorf("failed getting Chrome stdout: %v", err)
+		}
+		readFromPipe := func(name string, reader io.ReadCloser) {
+			r := bufio.NewReader(reader)
+			defer reader.Close()
```
gulien:
```suggestion
			defer reader.Close() // nolint: errcheck
```

build/base/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:9.5-slim
+FROM debian:9-slim
```
gulien: 👍

build/base/Dockerfile
```diff
@@ -91,3 +91,9 @@ RUN apt-get install -y \
     fonts-unfonts-core
 
 COPY build/base/fonts.conf /etc/fonts/conf.d/100-gotenberg.conf
+
+RUN groupadd --gid 1001 gotenberg \
```
gulien:
```suggestion
# |--------------------------------------------------------------------------
# | Default user
# |--------------------------------------------------------------------------
# |
# | TODO find a correct description for why we're doing this.
# |

RUN groupadd --gid 1001 gotenberg \
```

internal/pkg/pm2/chrome.go
```diff
@@ -58,12 +62,20 @@ func (p *chrome) viable() bool {
 	// check if Google Chrome is correctly running.
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
-	_, err := devtool.New("http://localhost:9222").Version(ctx)
-	return err == nil
+	p.manager.notifyf(`%v: checking Chrome liveness via debug version endpoint 
+'http://localhost:9222/json/version'`, p.name())
+	v, err := devtool.New("http://localhost:9222").Version(ctx)
+	if err != nil {
+		p.manager.notifyf("%v: Chrome version endpoint returned error: %v", p.name(), err)
```
gulien:
```suggestion
		p.manager.notifyf("%s: %s version endpoint returned error: %v", p.name(), p.Fullname(), err)
```

internal/pkg/pm2/chrome.go
```diff
@@ -58,12 +62,20 @@ func (p *chrome) viable() bool {
...
+		return false
+	}
+	p.manager.notifyf("%v: Chrome returned version info: %+v", p.name(), *v)
```
gulien:
```suggestion
	p.manager.notifyf("%s: %s returned version info: %+v", p.name(), p.Fullname(), *v)
```

internal/pkg/pm2/chrome.go
```diff
@@ -58,12 +62,20 @@ func (p *chrome) viable() bool {
...
+	return true
 }
 
 func (p *chrome) warmup() {
-	time.Sleep(5 * time.Second)
+	p.manager.notifyf("%v: allowing Chrome %v to startup", p.name(), warmupTime)
```
gulien:
```suggestion
	p.manager.notifyf("%s: allowing %s %v to startup", p.name(), p.Fullname(), warmupTime)
```

internal/pkg/pm2/pm2.go
```diff
@@ -76,8 +83,43 @@ func (m *processManager) pm2(p Process, cmdName string) error {
 		"pm2",
 		cmdArgs...,
 	)
+	m.notifyf("executing command '%v'", strings.Join(cmd.Args, " "))
+	if m.verbose {
+		chromeStdErr, err := cmd.StderrPipe()
```
gulien:
```suggestion
		processStdErr, err := cmd.StderrPipe()
```

It may also be `unoconv` 😄

build/package/Dockerfile
tbflw: No, it's not *really* necessary - just a result of our own habit of never copying more than is absolutely necessary.

tbflw: No, it's not *really* necessary - just a result of our own habit of never copying more than is absolutely necessary to facilitate best use of the Docker cache.

gulien:
@tbflw sorry for the delay! I'll try to merge your PR next week, I'm quit busy atm. 👍

PR #100, opened 2019-07-24
Noto emoji font

vbaidak:
This pull request is about adding missing emoji font.

It will replace this one:
image

with this one:
image

Please note, that **ttf-ancient-fonts** could be also installed via apt. However in that case, emojis won't be colored


gulien:
Hello @vbaidak! Are those "standard" emojis? Or this is just for the font displayed in your PR message?

vbaidak:
Hi @gulien .

This font supports all emojis defined in Unicode (v10) standard. See [this](https://www.google.com/get/noto/help/emoji/).

So, it should work for all emojis.

See also:
https://emojipedia.org/
https://emojipedia.org/white-medium-star/

gulien:
Looks great, thank you :)

gulien:
Thank you 👍

PR #109, opened 2019-08-31
Update index.html

smatyas:
removed some unprintable chars (hex '1c') rendered as '' from the doc


gulien:
Hello @smatyas 

Actually the documentation is generated from markdown files using `make doc`.

PR #114, opened 2019-09-24
WIP: Add Python example documentation

eamanu:
Add examples about the use of Gotenberg via Python

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -346,6 +346,20 @@ $dirPath = &#34;/foo&#34;;
 $filename = $client-&gt;store($request, $dirPath);
 </pre>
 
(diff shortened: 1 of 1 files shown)


gulien:
Hello @eamanu 

Thank you for this PR. Currently, I'd like to have only examples using `cURL` or the libraries.

Maybe you could create a `gist` or a repository with complete `python` examples? I could then provide a link in the documentation :)

eamanu:
Hi @gulien Thanks for you response.

Sadly I not complete the PR yet, but I can make a gist for that. 

>  I could then provide a link in the documentation :) 
That would be great.

Closing PR

PR #119, opened 2019-10-02
Correct A4 page height

foladipo:
The dimensions of an A4 sheet of paper are 8.27in x 11.69in. So, if the margins on the top and bottom of the page are both set to zero, the page height will be 11.69in, not 11.27in.


gulien:
Good catch! You should actually update the file located at https://github.com/thecodingmachine/gotenberg/blob/master/build/docs/content/04-html.md and then run `make doc` 👍

foladipo:
Roger that @gulien. I see you've implemented the fix yourself in one of your PRs. So it's all good, and I'm glad to be of help.

PR #128, opened 2019-10-12
Add basic authentication

odarriba:
**Summary**

This MR adds some changes we have been using internally, in order to add HTTP Basic authentication out-of-the-box.

We know that project's idea about this is that authentication should be provided by external software, but this can be useful in some scenarios, like running on Google Cloud Run.

This PR fixes/implements the following **features**

* [x] HTTP Basic authentication

The main motivation is, as explained above, allowing the container to run standalone in a serverless environment like the ones provided on AWS or GCP.

**Test plan (required)**

I have added tests that cover the functionality and this behaviour is configurable by using environment variables.

**Closing issues**

It does not close, but can help to the people watching [this issue](https://github.com/thecodingmachine/gotenberg/issues/90)

**Checklist**

- [x] Have you followed the guidelines in our [CONTRIBUTING](CONTRIBUTING.md) guide?
- [x] Have you written new tests for your core changes, as applicable?
- [x] Have you updated the documentation (`make doc`)?
- [x] This change has comments for package types, values, functions, and non-obvious lines of code

### DIFF
--- a/build/package/Dockerfile
+++ b/build/package/Dockerfile
@@ -10,8 +10,8 @@ FROM thecodingmachine/gotenberg:workspace AS workspace
 ARG VERSION
 
 ENV GOOS=linux \
-    GOARCH=amd64 \
-    CGO_ENABLED=0
+  GOARCH=amd64 \
+  CGO_ENABLED=0
 
 # Define our workding outside of $GOPATH (we're using go modules).
 WORKDIR /gotenberg/package
@@ -20,7 +20,7 @@ WORKDIR /gotenberg/package
 COPY go.mod go.sum ./
 
 RUN go mod download &&\
-    go mod verify
+  go mod verify
 
 # Copy our source code.
 COPY internal ./internal
@@ -42,9 +42,9 @@ LABEL authors="Julien Neuhart <j.neuhart@thecodingmachine.com>"
 
 ARG TINI_VERSION
 
-ADD https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini /tini
-RUN chmod +x /tini
-ENTRYPOINT [ "/tini", "--" ]
+ADD https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini-static /usr/local/bin/tini
+RUN chmod +x /usr/local/bin/tini
+ENTRYPOINT [ "tini", "--" ]
 
 # |--------------------------------------------------------------------------
 # | Final touch
--- a/internal/app/xhttp/xhttp.go
+++ b/internal/app/xhttp/xhttp.go
@@ -12,7 +12,7 @@ func New(config conf.Config) *echo.Echo {
 	srv.HideBanner = true
 	srv.HidePort = true
 
-	if (config.EnableAuthentication()) {
+	if config.EnableAuthentication() {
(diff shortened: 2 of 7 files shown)


internal/pkg/conf/conf_test.go
```diff
@@ -224,6 +224,78 @@ func TestDefaultListenPortFromEnv(t *testing.T) {
...
+	assert.Equal(t, expected, result)
+	os.Unsetenv(EnableAuthEnvVar)
+	os.Setenv(EnableAuthEnvVar, "0")
+	expected = DefaultConfig()
+	expected.enableAuthentication = false
+	result, err = FromEnv()
+	assert.Nil(t, err)
+	assert.Equal(t, expected, result)
+	os.Unsetenv(EnableAuthEnvVar)
+	// DISABLE_GOOGLE_CHROME wrongly set.
```
gulien: Typo 😄

gulien:
👍 thank you for this PR

internal/pkg/conf/conf_test.go
odarriba: Whoops! Copy-paste issue 🙈 Fixed!

build/package/Dockerfile
```diff
@@ -42,9 +42,9 @@ LABEL authors="Julien Neuhart <j.neuhart@thecodingmachine.com>"
 
 ARG TINI_VERSION
 
-ADD https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini /tini
-RUN chmod +x /tini
-ENTRYPOINT [ "/tini", "--" ]
+ADD https://github.com/krallin/tini/releases/download/v${TINI_VERSION}/tini-static /usr/local/bin/tini
```
gulien: What is the difference between `tini` and `tini-static`?

odarriba: It was a try to solve Cloud Run issue, related to this: https://github.com/krallin/tini#statically-linked-version

But it didn't work. Although it is safer to use statically linked binaries, is you want I can revert this change!

gulien: Alright, let's keep it this way then 👍

internal/app/xhttp/xhttp.go
```diff
@@ -10,10 +11,21 @@ func New(config conf.Config) *echo.Echo {
 	srv := echo.New()
 	srv.HideBanner = true
 	srv.HidePort = true
+
+	if config.EnableAuthentication() {
+		srv.Use(middleware.BasicAuth(func(username, password string, c echo.Context) (bool, error) {
+			if username == config.AuthenticationUsername() && password == config.AuthenticationPassword() {
```
bgentry: FYI @odarriba @gulien at a glance it appears that this is vulnerable to timing attacks because you're using plain `==` to compare credentials rather than using a secure constant time comparison operation.

odarriba: You are right. In theory, if the strings to compare are strong and the network conditions remains stable, one could really measure when a username is really valid.

This scenario is more feasible when the password check is more time consuming and we don't do it if the username doesn't match. However, it's better to prevent that situation by fixing this.

Really liked the example of `subtle.ConstantTimeCompare`. If this is reconsidered again, a look should be taken to that implementation.

Thanks for the review <3

PR #136, opened 2019-10-31
6.1.0

gulien:


gulien:
@odarriba for the authentication, best option would not be using an IAM or else? See https://cloud.google.com/run/docs/authenticating/overview.

odarriba:
@gulien It depends on the use case. In general, if you are not running all your services under GCP, the authentication flow is a bit over overwhelming to implement, specially on some technologies. 

That converts a pretty simple implementation of a Gotenberg client in a complex authorization behaviour previous to the call.

I was concerned about adding such a simple authentication because in general this can be got putting NGiNX (or any other reverse proxy) in front of Gotenberg, and control the authentication phase there. 

And of course I really understand the idea of keeping Gotenberg as simple as possible (or adding just the functionality really needed for the service).

If the matter is about adding this kind of _not so general_ options to public gotenberg image, please feel free to revert the merge, as we ca continue maintaining our private fork with these changes.

We really admire this software, and respect all decissions of the maintainers ❤️

PR #155, opened 2019-12-06
6.1.0

### LINKED ISSUE #81: Authentification to the remote URL
Is there a way to add an authentification key to the header of the request send by the gotenberg API when it fetch the web page specified in the remoteURL parameter ?

I need to create a pdf from an app page that isn't publicly accessible, so I'd like to authorize the gotenberg API to access, without adding the authentification in the URL, that wouldn't be very secured.

Is there a way to do that that I missed in the documentation, or is anything like this planned for a future version ?

### LINKED ISSUE #84: Whether the pagerange parameter can be added to specify the print range
I have a requirement for preview, which will get stuck when converting large files. I hope you can add such a parameter to speed up the conversion. Thank you

### LINKED ISSUE #116: Can't ask Gotenberg to fetch url from docker with multi-domain app
Hello folks!

I've just discovered Gotenberg and I configured it in my `docker-compose.yaml` file, the exposed port is okay, and I can generate PDFs from remote URLs.

However, it only works with _remote_ urls.

Let me explain.

I have a PHP application which **must** be multidomain.

Let's consider such Compose config:

```yaml
services
    app:
        # It exposes port 8000, and requests on "wrong" domains such as 127.0.0.1 return HTTP 400 error

    gotenberg:
        image: thecodingmachine/gotenberg:5
```

I'll need to generate PDFs from the `app` container, therefore I will execute something like this from inside this container:

```
$ curl --request POST \
    --url http://gotenberg:3000/convert/url \
    --header "Content-Type: multipart/form-data" \
    --form remoteURL=http://mydomain.docker:8000/something.html \
    -o result.pdf
```

However, this will not work: `gotenberg`'s container doesn't know anything about the `mydomain.docker` host, so I end up having a blank page in `result.pdf`.
By the way, any `curl` call to the same address will result in a `curl: (7) Failed to connect to mydomain.docker port 8000: Connection refused` error.

I even tried `echo "mydomain.docker  app" >> /etc/hosts`, without success (doesn't seem to be taken in account at all).

I could find a hack that would have worked with Curl, by "mocking" the host:

```
$ curl http://app:8000/ -H"Host: mydomain.docker"
```

Using this trick, `curl` will effectively call the `app` host (the same as the app's container name, valid inside a Docker network), but the resolved host will be `mydomain.docker`. It just works with `curl`.

However, it does not go beyond as this trick does not seem to be possible with Gotenberg.

If it was possible to use Gotenberg without a separate Docker container, it would probably be easier, as I could directly install it in the `app` container and make all queries "from inside" (it would make the image heavier but would be worth it IMO). But I'm not sure copy/pasting config and setup from your repository would be worth it nor maintainable.

I also tried using the `/html` endpoint, but there are so many CSS, JS and images in the web page that I don't even know how I can achieve that without hundreds of lines of code full of hacks or overengineered stuff.

What I'm almost sure about is that it would _probably_ work in production. I don't use any "complex" setup (just have one dedicated server) so running the docker image alongside all other apps seems pretty okay, and as it would call totally remote addresses (with proper DNS-based hosts), it would work without harm. But for dev, it's not possible.

I'm starting to think that I won't be able to use Gotenberg at all with the setup I have...

Do you have any idea of how I could solve this?

As a side-note, I have chromium installed on the main Docker image (necessary to use Panther for testing).
Should I look further there?

### LINKED ISSUE #125: Assign another uid to user
Would it be possible for me to specify another uid for the user `gotenberg`? By default it's `1001`, but certain environments have constraints on the range of available uids for security reasons.

### LINKED ISSUE #152: Allow http namespacing in the url
Your issue may already be reported!
Please search on the [issue tracker](../../../issues) before creating one.

## Expected Behavior

I would like the ability to set the root url path for the service so that it doesn't pollute the / path on the load balancer. something like `https://domain.com/gotenberg/convert/html` would be helpful since I do service discovery via url paths.

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Version used:
* Operating System and version:
* Link to your project:

gulien:
#### Custom root path

You may now customize the root path of the API using the environment variable `ROOT_PATH` (default to `/`).

This is useful if you wish to do service discovery via URL paths.

Fixes #152.

#### Custom HTTP headers

You may now specify your own HTTP headers for your:

* URL conversions.
* conversions with a webhook.

Fixes #81.
Fixes #116.

#### Page ranges

You may now specify a page ranges for HTML, Markdown, URL and Office conversions.

Limitation: does not work for `/merge` endpoint

Fixes #84.

#### Allowing custom UID and GID for Gotenberg user

One may now specify a custom UID and GID thanks to two new build arguments:

* `GOTENBERG_USER_GID`
* `GOTENBERG_USER_UID`

**Note:** it is not recommended to do it on runtime in production (see https://github.com/boxboat/fixuid) that's why you'll have to re-build the Docker image.

Fixes #125.


Pierstoval:
Thanks a LOT for this release! 👍 🚢 🚀

PR #182, opened 2020-03-18
Update index.html

sdawka:
**Summary**

Changes to the docs to prevent misinformation

Explain the **motivation** for making this change. What existing problem does the pull request solve?

The cloud run docs clearly says that async activity is not to be done, and I found out the hard way, so I'd like to save other's time.


gulien:
Hello @sdawka 

Thanks for the PR! You should actually updates the Markdown files under `build > docs > content` and then run `make doc`

> The cloud run docs clearly says that async activity is not to be done, and I found out the hard way, so I'd like to save other's time.

Good to know, I was not aware of that limitation. I guess they want to prevent DDOS capabilities 🤔

PR #204, opened 2020-06-01
Added warning for URLRequest and service named app

WillyReyno:
A docker-compose service named "app" will systematically return a blank PDF when using `URLRequest()`.
Added a warning to recommend people to rename their service if it's named "app".


gulien:
Hello @WillyReyno 

Thanks for the PR! 

You should actually update the markdown file located at 'build/docs/content' and then run `make doc`

gulien:
ping @WillyReyno 😄

PR #241, opened 2020-09-17
[bugfix/issue-172] fixing health check

Tealsky:
**Summary**

This PR fixes/implements the following **bugs/features**

* [x] When an internal condition appears to be failing, the healthcheck returns healthy. 

**Test plan (required)**
image

image

**Closing issues**

Fixes https://github.com/thecodingmachine/gotenberg/issues/172

**Checklist**

- [x] Have you followed the guidelines in our [CONTRIBUTING](CONTRIBUTING.md) guide?
- [ ] Have you written new tests for your core changes, as applicable?
- [ ] Have you updated the documentation (Markdown files under `build > docs > content` and then `make doc`)?
- [ ] This change has comments for package types, values, functions, and non-obvious lines of code

### DIFF
--- a/internal/pkg/chrome/chrome.go
+++ b/internal/pkg/chrome/chrome.go
@@ -31,7 +31,8 @@ func Start(logger xlog.Logger, ignoreCertificateErrors bool) error {
 		}
 		// if the process failed to start correctly,
 		// we have to restart it.
-		if !isViable(logger) {
+		isViable, _ := IsViable(logger)
+		if !isViable {
 			return restart(logger, cmd.Process, ignoreCertificateErrors)
 		}
 		return nil
@@ -117,7 +118,8 @@ func restart(logger xlog.Logger, proc *os.Process, ignoreCertificateErrors bool)
 		}
 		// if the process failed to restart correctly,
 		// we have to restart it again.
-		if !isViable(logger) {
+		isViable, _ := IsViable(logger)
+		if !isViable {
 			return restart(logger, cmd.Process, ignoreCertificateErrors)
 		}
 		return nil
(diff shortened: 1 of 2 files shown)


gulien:
Thank you @Tealsky for this PR :smile:

I'll take a look at it 👍

Tealsky:
> Thank you @Tealsky for this PR 
> 
> I'll take a look at it

No problem :) But I want to create a test by using mock, but there isn't any mock package in the project. Is it recommended to use mock or not ? Sorry, I'm still new in the Go world.

gulien:
Looks good for me 👍 I'm just wondering if we should restart Google Chrome when a probe call the healthcheck endpoint 🤔 
Did you taste the behaviour "live"?

Concerning tests, currently it is a bit of a mess, as we are testing each Go files individually. I plan to work on a simple solution for v7.

Tealsky:
> Looks good for me  I'm just wondering if we should restart Google Chrome when a probe call the healthcheck endpoint 
> Did you test this behaviour "live"?
> 
> Concerning tests, currently it is a bit of a mess, as we are testing each Go files individually. I plan to work on a simple solution for v7.

Yes, I have tested the behavior by killing the Google Chrome process, in the docker container (ps -ef | grep 9222).
image
And, I'm using Kubernetes to deploy Gotenberg. In that case I can configure a liveness probe that will restart the container if the healthcheck is unhealthy. It's working perfectly.
image

image

Concerning the tests, I am a little annoyed to reduce the code coverage :(

internal/pkg/chrome/chrome.go
```diff
@@ -149,21 +152,23 @@ func isViable(logger xlog.Logger) bool {
...
 		logger.DebugOpf(
 			op,
 			"Google Chrome headless is viable as endpoint returned '%v'",
 			v,
 		)
-		return true
+		return true, nil
 	}
 	result := false
+	err := error(nil)
```
gulien, line 165:
```suggestion
	err := nil
```

gulien:
Actually my bad, I read your PR too fast: you're just checking that Google Chrome is viable, but you do not try to restart it directly from the the ping handler. Perfect 😄 

I've added a minor suggestion, as soon as you update this line I can create a new release 👍 

> Concerning the tests, I am a little annoyed to reduce the code coverage :(

As v7 code will be drastically different from current version, that's OK 😉

internal/pkg/chrome/chrome.go
Tealsky: image

Unfortunately, the short version doesn't work but the long version seems to be ok.

```suggestion
	var err error
```

gulien: My bad, `var err error` is the right way 👍

gulien:
Thank you very much @Tealsky 🎉

PR #312, opened 2021-07-02
Dockerfile: Use https everywhere

j2L4e:
Changed plain http downloads to https ones.

If it actually IS safe to download these files via plain http, please pardon my ignorance. An explanation would be appreciated. I'm not familiar with the inner workings of `.deb`s / apt, but this doesn't seem right to me.


gulien:
@j2L4e No issue on the CI, so I guess we're good 👍

PR #313, opened 2021-07-04
Feature add arm support and use chromium

toxix:
This allows to build and run gotenberg on a raspbarry 4b.

We are using chromium as there is no arm build of chrome.

The tiny executable needs to be changed to the version that is
compatible with the correct architecture.

Also preparing for multi-architecture building of the docker images in
the publish script.

We removed the crosscompiling enviornment variables of go, as they are
preventing a compiling for arm and they will be set automatically.

fixes https://github.com/thecodingmachine/gotenberg/issues/213

**Test plan**
Please check if this changes are not braking anything. I have tested the code only on an Raspbarry Pi v4

### DIFF
--- a/build/base/Dockerfile
+++ b/build/base/Dockerfile
@@ -1,4 +1,5 @@
-FROM debian:buster-slim
+ARG ARCH=
+FROM ${ARCH}debian:buster-slim
 
 # |--------------------------------------------------------------------------
 # | Common libraries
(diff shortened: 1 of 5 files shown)


blackfurball:
This PR is a godsend. 
Im using Gotenberg alongside https://github.com/jonaswinkler/paperless-ng and Apache Tika.
The latte two running easily on my Pi4B. Just Gotenberg needs to run on x86 host atm. 

Looking forward to a Pi4 compatible arm release :)

gulien:
Hello @toxix :) Thanks for your PR! 

Alas, I don't plan to support arm64 architecture for Gotenberg 6. 
That being said, I'm currently working on the next version of Gotenberg, and it will bring, among other things, arm64 support :)

kirosc:
The change works on my Apple M1 👍🏻

PR #314, opened 2021-07-09
Add tags on openapi.yaml endpoints

AlexBondor:
**Summary**

This PR fixes/implements the following **features** - this is actually a minor fix (nice to have)

* [ ] On generating the Java client using openapi.yaml the tag will be used to name the Java class which will provide the client implementation

Explain the **motivation** for making this change. What existing problem does the pull request solve?

I wanna use gotenberg as a docker service and make requests from my spring-boot app. I didn't want to write an http client on my own as I saw that you provide an openapi.yaml spec. Thus, I'm using an openapi generator in my application, feeding your app's openapi spec and voila: The java client that I could already use gets automatically generated. The drawback was that the endpoints in the openapi spec didn't provide any `tags` and then the default name for the Java client would have been `DefaultApi`.. that's not intuitive at all. (see the snapshots below)

**BEFORE**
image

**AFTER**
image

### DIFF
--- a/docs/openapi.yaml
+++ b/docs/openapi.yaml
@@ -141,7 +141,7 @@ paths:
   /convert/markdown:
     post:
       tags:
-        - Html2Pdf
+        - Gotenberg
       summary: Convert a Markdown file to PDF
       description: >-
         Send an HTML file called `index.html` as a multipart form request and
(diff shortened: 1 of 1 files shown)


gulien:
Thanks @AlexBondor!

PR #322, opened 2021-08-24
Update LibreOffice to 7.2 stable

mrcageman:
Replaced the `buster-backports` version of Libreoffice 7.0 with the 7.2 stable version using the `.deb` from the official website.

Reason for this version update is that I ran into multiple issues where DOCX files with headers/footers would not convert to PDF with LibreOffice 7.0. Updating to 7.1 fixed this issue for me. Since 7.2 was marked stable, I felt updating to the latest stable was the proper direction. 

If this is not the direction you wish to go, then I hope this PR is still useful to those who've run into similar issues.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -102,11 +102,15 @@ RUN \
     fonts-sil-gentium \
     fonts-sil-gentium-basic &&\
     rm -f ./ttf-mscorefonts-installer_3.8_all.deb &&\
-    echo "deb https://httpredir.debian.org/debian/ buster-backports main contrib non-free" >> /etc/apt/sources.list &&\
+    # See https://github.com/nextcloud/docker/issues/380.
+    mkdir -p /usr/share/man/man1mkdir -p /usr/share/man/man1 &&\
+    curl -L "https://download.documentfoundation.org/libreoffice/stable/7.2.0/deb/x86_64/LibreOffice_7.2.0_Linux_x86-64_deb.tar.gz" | tar -xz &&\
     apt-get update -qq &&\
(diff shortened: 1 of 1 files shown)


build/Dockerfile
```diff
@@ -102,11 +102,14 @@ RUN \
     fonts-sil-gentium \
     fonts-sil-gentium-basic &&\
     rm -f ./ttf-mscorefonts-installer_3.8_all.deb &&\
-    echo "deb https://httpredir.debian.org/debian/ buster-backports main contrib non-free" >> /etc/apt/sources.list &&\
+    # See https://github.com/nextcloud/docker/issues/380.
+    mkdir -p /usr/share/man/man1mkdir -p /usr/share/man/man1 &&\
+    curl -L "https://download.documentfoundation.org/libreoffice/stable/7.2.0/deb/x86_64/LibreOffice_7.2.0_Linux_x86-64_deb.tar.gz" | tar -xz &&\
```
gulien, line 107: It will certainly work on `amd64` architectures, but for not for `arch64` ones.

Note the `x86_64` part of the URL 😄

gulien:
Hello @mrcageman!

Thanks for this PR 👍 We need to find a way to make it work on both `amd64` and `arch64` architectures 🤔

mrcageman:
I don't know how this works with the package from the `buster-backports` repository, but as far as I'm aware LibreOffice itself only offers `amd64` builds. I myself have no way to test this on an `arch64` machine, but if anyone else does it would be helpful in figuring this out.

mrcageman:
Thanks for the update, much appreciated! Happy I could be of any help to you :)

PR #340, opened 2021-09-08
Add supported file extensions for unovonv

remoteexception:
Added the following file extensions for unoconv:
- tif
- jpeg
- odp


pkg/modules/libreoffice/unoconv/unoconv.go
```diff
@@ -243,9 +244,11 @@ func (mod Unoconv) Extensions() []string {
 		".swf",
 		".sxd",
 		".sxw",
+		".tif",
```
gulien, line 247: I didn't find `.tif` in the list of supported file extensions (`unoconv --show`) 🤔  `.tif` == `.tiff` ?

remoteexception: Yes you are right. But unoconv actually supports .tif.

remoteexception: jpeg isn't in the list too, but is supported by unoconv btw.

gulien (approved):

gulien:
Thanks @remoteexception 😄

PR #341, opened 2021-09-08
Fix/use unrestricted webhook ports

remoteexception:
Use ports between 1025 and 65536 to prevent operation system error. In some OS only root is allowed to use ports between 0 and 1024.


gulien:
Thanks @remoteexception for this MR!

Could you remove the commits related to #340? 😄

pkg/modules/api/middlewares_test.go
```diff
@@ -111,7 +111,7 @@ func TestHttpErrorHandler(t *testing.T) {
 
 		func() {
 			rand.Seed(time.Now().UnixNano())
-			webhookPort := rand.Intn(65535-1+1) + 1
+			webhookPort := rand.Intn(64511-1+1) + 1025
```
gulien, line 114:
```suggestion
			webhookPort := rand.Intn(65535 - 1025 + 1) + 1025
```

pkg/modules/api/middlewares_test.go
```diff
@@ -825,7 +825,7 @@ func TestContextMiddlewareWithWebhook(t *testing.T) {
 			webhook.HidePort = true
 
 			rand.Seed(time.Now().UnixNano())
-			webhookPort := rand.Intn(65535-1+1) + 1
+			webhookPort := rand.Intn(64511-1+1) + 1025
```
gulien, line 828:
```suggestion
			webhookPort := rand.Intn(65535 - 1025 + 1) + 1025
```

gulien (changes requested):

PR #346, opened 2021-09-09
Only use webhook ports between 1025 and 65536 in tests

remoteexception:
- ports between 0 and 1024 can only be used by root in some OS


remoteexception:
Sorry Julian, I accedentially removed my branch. Here is the new one, already with your improvement.

gulien:
Thansk @remoteexception 😄

PR #347, opened 2021-09-09
fix: use the correct dimensions of A4 paper size

remoteexception:
The correct paper dimensions for A4 are:
- width: 8.3 inch
- height: 11.7 inch
This is the ISO 216 standard. See https://www.papersizes.org/a-paper-sizes.htm


gulien:
👋 @remoteexception 

Thanks for your interest in Gotenberg :) Regarding the A4 paper dimensions: Gotenberg uses by default the `Letter` format. If you want to change the dimensions of the resulting PDF, you should take a look at the `paperWidth ` and `paperHeight` form fields (see https://gotenberg.dev/docs/modules/chromium#routes).

remoteexception:
But why don't you set the dimensions as specified in the ISO-216 standard? As it is now, you always have to send the parameters to enforce the standard.

remoteexception:
I wondered why you changed the dimensions from version 6 to 7. In version 6 it fit better the A4 paper than it does now.

remoteexception:
Sorry Julien, the "Letter" format is a more common paper format then A4? I'm not familiar with paper formats. I thought the given sizes are for A4. If not, then forget my comments.

gulien:
The `Letter` format is actually the default paper size from the Chrome DevTools Protocol (~ the API I use for interacting with Chromium/Google Chrome).  So, for v7, I decided to be as close as possible to the default values, including the paper size 😄 

You are still able to customize the paper size. For instance, if you want to enforce the A4 format:

```bash
curl \
--request POST 'http://localhost:3000/forms/chromium/convert/url' \
--form 'url="https://my.url"' \
--form 'paperWidth="8.27"' \
--form 'paperHeight="11.7"'
```

Of course, as you said, you know have to send the related form fields each time you call Gotenberg.

Anyway, consider creating an issue with your arguments 😄 maybe it will gain some tractions by other users and help me make a decision 👍

Issue #365, opened 2021-09-24
Gotenberg over HTTP/2


Svendegroote91:
### Discussed in https://github.com/gotenberg/gotenberg/discussions/364

<div type='discussions-op-text'>

<sup>Originally posted by **Svendegroote91** September 23, 2021</sup>
Currently we are facing 413 responses from Cloud Run when we invoke Gotenberg over HTTP/1. 
I believe HTTP/2 is currently not supported, is this on the roadmap?</div>

gulien:
@Svendegroote91 Could you test the image `gulnap/gotenberg:http2-cloudrun` on CloudRun ? It supports HTTP2 (H2C) 😄 (thanks to @KillianMeersman). If it works as expected, I'll release a new version of Gotenberg 🚀

Svendegroote91:
> @Svendegroote91 Could you test the image `gulnap/gotenberg:http2-cloudrun` on CloudRun ? It supports HTTP2 (H2C) 😄 (thanks to @KillianMeersman). If it works as expected, I'll release a new version of Gotenberg 🚀

Awesome! We'll try to test that this week, keep you posted 👍

gulien:
@Svendegroote91 Did you find some time to test it? 😄

Svendegroote91:
> @Svendegroote91 Did you find some time to test it? 😄

Yes, works like a charm 👌

gulien:
Great new! Thanks 😊

PR #366, opened 2021-09-24
Add h2c support

KillianMeersman:
Add support for HTTP/2 cleartext (h2c) via a new flag: `--api-http-version`, with possible values of `http1` (default) and `h2c`.
This is needed when running on Cloud Run, as HTTP/1.1 requests are limited to a maximum upload size of 32MB.

It seems that Echo's HTTP/2 server also supports HTTP/1.1, it should be possible to remove the original `a.srv.Start(address)` call and support both. I left both in just to be safe.

### DIFF
--- a/pkg/modules/api/api.go
+++ b/pkg/modules/api/api.go
@@ -26,13 +26,6 @@ func init() {
 	gotenberg.MustRegisterModule(API{})
 }
 
-type HTTPVersion uint
-
-const (
-	HTTP1 HTTPVersion = iota
-	H2C   HTTPVersion = iota
-)
-
 // API is a module which provides an HTTP server. Other modules may add
 // "multipart/form-data" routes, middlewares or health checks.
 type API struct {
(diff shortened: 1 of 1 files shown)


gulien:
Thanks a lot @KillianMeersman! I did not expect to have a PR on HTTP2 so fast lol 🚀 

I wonder if we could just enable HTTP2 by default, i.e., no dedicated flag and so on.

gulien:
Closes #365.

KillianMeersman:
> Thanks a lot @KillianMeersman! I did not expect to have a PR on HTTP2 so fast lol rocket
> 
> I wonder if we could just enable HTTP2 by default, i.e., no dedicated flag and so on.

Should be possible, I'll experiment a bit and let you know.

KillianMeersman:
Experimented with always supporting HTTP/2 and it seems to work well with both HTTP1.1 and HTTP/2 clients. I pushed a new change removing the flag and using the h2c server.

gulien:
Thanks @KillianMeersman 👍

PR #374, opened 2021-10-22
feat: add exiftool module, integrate with Libreoffice conversion

padpadthai:
Creates an [Exiftool](https://exiftool.org/) module for writing file metadata (particularly relevant to PDF in this case).

The Exiftool module is integrated with the LibreOffice module and its PDF conversion route, i.e. `/forms/libreoffice/convert`. Metadata can be supplied to this endpoint in JSON format via the `metadata` from data element. I can help update the docs repository should this PR be merged. A sample request

`curl --location --request POST 'localhost:3000/forms/libreoffice/convert' --form 'files=@"/tmp/test-large.gif"' --form 'metadata="{\"Creator\": \"Personio\", \"Producer\": \"Workzag\" }"'`

Limited JSON metadata datatypes are currently supported, i.e. string and number. However, there is potential to support additional data types in the future.

I tried to follow existing patterns in the project. Let me know if you have any comments, opinions or feedback :)


gulien:
Hello @padpadthai 👋 

Thank you very much for your time on this PR 👍 , and sorry for the delay!

I do think the `exiftool` module should be a PDF Engine (i.e., https://github.com/gotenberg/gotenberg/blob/v7.1.0/pkg/gotenberg/pdfengine.go#L34).

You could add a method `WriteMetadata` to the `PDFEngine` interface, for instance:

```go
WriteMetadata(ctx context.Context, logger *zap.Logger, metadata map[string]interface{}, inputPath string) error
```

Then I guess the path is the following:

1. Make the `exiftool` module a PDF Engine (see for instance https://github.com/gotenberg/gotenberg/blob/v7.1.0/pkg/modules/pdftk/pdftk.go).
2. Update other PDF Engine modules with the new `WriteMetadata` method (without support).
3. Update the `pdfengines` module (https://github.com/gotenberg/gotenberg/blob/v7.1.0/pkg/modules/pdfengines/multi.go).
4. Update the `chromium`, `libreoffice`, and `pdfengines` routes.

Once there, we could start improving the implementation to support more datatypes 😄

padpadthai:
Thanks for the feedback @gulien .

I should get some time to refactor based on your outline soon :slightly_smiling_face:.

Also on another note, I see from another issue that you're now a father. Big congatulations :baby:

Issue #287, opened 2021-03-08
unoconv GPL license


pawepaw:
Hi,

How is it possible that you're using `unoconv` which has GPL-2 license and you're exposing `gotenberg` as MIT? Do you have some special agreement with `unoconv` or am i missing something?

gulien:
Hello @pawepaw, no I don't. 

@dagwieers or @regebro: if there is an issue on that matter, feel free to contact me 😃

pawepaw:
Well wanted to use it but the GPL license disqualifies my usage plan, probably more users would have the same problem if they spot that they are actually using GPL soft, not MIT.

Would be good to clarify it.

dagwieers:
IANAL. I don't think there is a problem. AFAICT unoconv is called as an external tool for performing the conversion.
There could have been issues if it would be calling unoconv as a library, but that is not the case here IMO.

gulien:
Thanks @dagwieers for the clarification 👍

pawepaw:
Well i'm not a lawyer as well :D thanks for fast answers!

PR #397, opened 2021-12-17
feat: add ghostcript module for conversion to PDF/A and merging

Vrex123:
Creates a [Ghostscript](https://ghostscript.com/) module for conversion to PDF/A-1b, PDF/A-2b, PDF/A-3b

Unoconv keeps messing fonts in my PDFs when converting HTML to PDF/A-1a, so I found Ghostscript, which doesn't have these problems.

Debian repositories should have older version of ghostscript, but I keep getting dpkg errors, when trying to install ghostscript from repository, so I had to download it from their's github.

It will be great to have some feedback :)

UPD: added support for merging PDF

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 /coverage.html
 /coverage.txt
+
+#Intellij IDEA project files
+.idea
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ GOTENBERG_USER_GID=1001
 GOTENBERG_USER_UID=1001
 NOTO_COLOR_EMOJI_VERSION=v2.028 # See https://github.com/googlefonts/noto-emoji/releases.
 PDFTK_VERSION=1527259628 # See https://gitlab.com/pdftk-java/pdftk/-/releases - Binary package.
+GHOSTSCRIPT_VERSION=9.55.0-linux-x86_64 # See https://github.com/ArtifexSoftware/ghostpdl-downloads/releases
 GOLANGCI_LINT_VERSION=v1.43.0 # See https://github.com/golangci/golangci-lint/releases.
 
 .PHONY: build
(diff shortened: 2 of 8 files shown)


gulien:
Hello @Vrex123!

Thanks for this PR 😄

I have a hard time understanding the [Ghostscript license](https://www.ghostscript.com/licensing/index.html). In short, I don't want to wrap a piece of software with a license that may conflict with their usage 🤔 Maybe @chris-liddell could help us?

Vrex123:
Fair point. If it'll harm Gotenberg, then it's not good idea to include it in Gorenber, yeah :thinking:

gulien:
But if this is not an issue, being able to convert PDFs to `PDF/A-1b`, `PDF/A-2b` and `PDF/A-3b` would be a killer feature 👍

Vrex123:
I tired to wrap my head around Ghostscript license. I even tried to write what I found out, but there are so many different opinions regarding GPL-related licenses.
Here, I am using Ghostscript binary if it's included in image, so Gotenberg source code is not linked to or depending on Ghostscript in any way.
Looks like there should be two separate Docker images: one with Ghostscript and end-user should agree to Ghostscript's license, and one without Ghostscript and nothing changing.

Or I am completely wrong, and should've pursued legal career to understand all of this clearly :dizzy_face:

gulien:
Thanks for the follow up @Vrex123 😄 

There was an issue (see #287) about the `unoconv` license and the author did not see any problem with our usage.

Regarding AGPL vs GPL, there might be a clause that can conflict with some users' usage, so I don't know 🤷‍♂️.

A solution in the meantime would be to create a variant of Gotenberg using the [template repository](https://github.com/gotenberg/gotenberg-template-repository).

Vrex123:
Finally had time to create [repository](https://github.com/Vrex123/gotenberg-ghostscript) with Ghostscript.

Guess you can close this PR now :smile:

gulien:
Thanks for your work @Vrex123!

On a side note, I find a way to add the `PDF/A-2b` and `PDF/A-3b` formats with `unoconv`. It might still break the fonts though.

Anyway, I'll add your repository to the awesome list 🚀

PR #399, opened 2021-12-27
feat: add Noto CJK fonts

fengkaijia:
`ttf-wqy-zenhei` is a community-maintained font that only includes a subset of CJK characters, and was last updated 10 years ago.

Noto Sans CJK and Noto Serif CJK (https://github.com/googlefonts/noto-cjk) by Google and Apple are the go-to fonts for compatibility now.


gulien:
Thanks @fengkaijia!

PR #428, opened 2022-02-15
New `forms/pdfengines/compress` endpoint

tito:
This endpoint allow to compress a pdf.
This can help a lot for chromium pdf export that can have large output.

*Back story*: we were having some issue (https://bugs.chromium.org/p/chromium/issues/detail?id=103602) and found that for one generated PDF from chromium/gotenberg was about 1.8MB. It can go down to 225KB by compressing it with ghostscript.

This include a new pdfengines/compress route, a new ghostscript pdfengine that implements it.

What do you think about a such new feature ?
Is is the right way to go ?

## Benchmark

All generated/compressed files gave us no visual difference on linux and windows/acrobat (mac not tested). We have an PDF named `initial.pdf`, and 3 tests:
```
$ ps2pdf initial.pdf output-1.pdf
$ ps2pdf -dPDFSETTINGS=/ebook initial.pdf output-2.pdf
$ gs -sDEVICE=pdfwrite -dCompatibilityLevel=1.4 -dPDFSETTINGS=/screen -dNOPAUSE -dQUIET -dBATCH -sOutputFile=output-3.pdf initial.pdf
$ ls -lah *.pdf
-rw-r--r-- 1 tito tito 1,8M 15 févr. 14:11 initial.pdf
-rw-r--r-- 1 tito tito 425K 15 févr. 14:12 output-1.pdf
-rw-r--r-- 1 tito tito 374K 15 févr. 14:12 output-2.pdf
-rw-r--r-- 1 tito tito 225K 15 févr. 14:13 output-3.pdf
```

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -138,6 +138,8 @@ RUN \
     DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends qpdf &&\
     # See https://github.com/nextcloud/docker/issues/380.
     mkdir -p /usr/share/man/man1mkdir -p /usr/share/man/man1 &&\
+    # Download ghostscript
+    DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends ghostscript &&\
     # Cleanup.
     # Note: the Debian image does automatically a clean after each install thanks to a hook.
     # Therefore, there is no need for apt-get clean.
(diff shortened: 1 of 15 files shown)


gulien:
Hello @tito 

Thanks for this well written PR 👍 

Alas, GhostScript licensing doesn't allow me to add it to the standard version of Gotenberg (see #397).

@Vrex123 also worked on a [GhostScript module](https://github.com/Vrex123/gotenberg-ghostscript), if you want to take a look 😄

gulien:
I'm not familiar with `ps2pdf`, and I cannot find it in the [Debian packages](https://packages.debian.org/search?suite=bullseye-updates&arch=any&searchon=names&keywords=ps2pdf) 😅

tito:
My bad, ps2pdf came from ghostscript as well. Then it's over, thanks for your consideration !

PR #429, opened 2022-02-16
fix(chromium): regex only accepts tmp folder by default

ubiuser:
The current deny list regex lets people use any permutation of [tmp] characters, so for example `file:///pmt/foo` is not denied.
Either add a simple regex to the allow list to lock access to the /tmp folder or update the default deny list to `^file:///(?:[^t]+|t(?:$|[^m]|m(?:$|[^p])))*$`.


gulien:
😄 *regex*...

gulien:
Thanks @ubiuser!

PR #454, opened 2022-05-02
Implement document thumbnailing

Jasonkingsmill:
This implements a new route at /forms/libreoffice/thumbnail for the purposes of "thumbnailing" the first page of a document format.
It takes a single imageFormat parameter, allowing conversion of a docx/pdf to jpg/tiff/png etc...

If you're willing to accept, I will also add some unit tests, or do some suggested refactoring

### DIFF
--- a/pkg/modules/libreoffice/libreoffice.go
+++ b/pkg/modules/libreoffice/libreoffice.go
@@ -75,6 +75,7 @@ func (mod LibreOffice) Routes() ([]api.Route, error) {
 
 	return []api.Route{
 		convertRoute(mod.unoAPI, mod.engine),
+		thumbnailRoute(mod.unoAPI, mod.engine),
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


gulien:
Hello @Jasonkingsmill 👋 

Thanks for your PR! The feature is really interesting, but I guess folks will want conversions to image formats for whole documents as well.

Right now, I'll prefer to stick to PDF only 😄

Jasonkingsmill:
Hi @gulien 
Would you accept this feature if it were to follow the existing pattern - using the same "/forms/libreoffice/convert" endpoint behavior, adding an additional option for the outputFormat as the image format (with zip result for image formats that don't support multi-page - Ie, anything other than tiff), or you still want to keep it as PDF only?

gulien:
Hello @Jasonkingsmill,

For now I still want to keep it as PDF only. The truth is that I don't have much time available for maintaining Gotenberg currently, so adding a whole new output format is too *risky* for me.

That being said, a lot of users have requested images as output format, so maybe in the future 😄

thaoula:
Hi @gulien,

Firstly, thanks for Gotenberg.

Just wondering if this could be packaged as a module or something we could include in a custom docker image? This way it would a plugin and not part of the Gotenberg you currently maintain?

Regards,
Tarek

gulien:
Hello @thaoula,

Definitely! Take a look at the https://github.com/gotenberg/gotenberg-template-repository 😄

PR #496, opened 2022-09-14
add encrypt route

vidschofelix:
First time even touching go. Implemented encryption of PDFs for qpdf and pdfcpu. Pdftk also supports encrypting, but behaviour was weird, so i dropped it for now. 

Api description was extended, but no tests so far, because i have no idea how they work in go. 

Please give feedback :)

### DIFF
--- a/pkg/gotenberg/pdfengine.go
+++ b/pkg/gotenberg/pdfengine.go
@@ -38,6 +38,9 @@ type PDFEngine interface {
 
 	// Convert converts the given PDF to a specific PDF format.
 	Convert(ctx context.Context, logger *zap.Logger, format, inputPath, outputPath string) error
+
+	//Encrypt one or more PDF Files with given passwords.
+	Encrypt(ctx context.Context, logger *zap.Logger, keyLength int, ownerPassword, userPassword, inputPath, outputPath string) error
 }
 
 // PDFEngineProvider is a module interface which exposes a method for creating a
--- a/pkg/modules/pdfengines/routes.go
+++ b/pkg/modules/pdfengines/routes.go
@@ -145,3 +145,57 @@ func convertRoute(engine gotenberg.PDFEngine) api.Route {
 		},
 	}
 }
+
+func encryptRoute(engine gotenberg.PDFEngine) api.Route {
+	return api.Route{
+		Method:      http.MethodPost,
+		Path:        "/forms/pdfengines/encrypt",
+		IsMultipart: true,
+		Handler: func(c echo.Context) error {
+			ctx := c.Get("context").(*api.Context)
+
+			// Let's get the data from the form and validate them.
+			var (
+				inputPaths  []string
+				ownerPassword string
+				userPassword string
+				keyLength int
+			)
+
+			err := ctx.FormData().
+				MandatoryPaths([]string{".pdf"}, &inputPaths).
+				MandatoryString("ownerPassword", &ownerPassword).
+				MandatoryString("userPassword", &userPassword).
+				Int("keyLength", &keyLength, 256).
+				Validate()
+
+			if err != nil {
+				return fmt.Errorf("validate form data: %w", err)
+			}
+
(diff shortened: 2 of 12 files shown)


pkg/gotenberg/pdfengine.go
```diff
@@ -38,6 +38,9 @@ type PDFEngine interface {
 
 	// Convert converts the given PDF to a specific PDF format.
 	Convert(ctx context.Context, logger *zap.Logger, format, inputPath, outputPath string) error
+
+	//Encrypt one or more PDF Files with given passwords.
```
gulien, line 42:
```suggestion
	// Encrypt encrypts the given PDF.
```

pkg/modules/pdfengines/routes.go
```diff
@@ -145,3 +145,57 @@ func convertRoute(engine gotenberg.PDFEngine) api.Route {
...
+				ownerPassword string
+				userPassword string
+				keyLength int
+			)
+
+			err := ctx.FormData().
+				MandatoryPaths([]string{".pdf"}, &inputPaths).
+				MandatoryString("ownerPassword", &ownerPassword).
+				MandatoryString("userPassword", &userPassword).
+				Int("keyLength", &keyLength, 256).
```
gulien, line 169: I think we have to validate if the value is either 40, 128 or 256.

gulien (commented):
Thanks for this PR! 😄 

Don't forget to run `make fmt`.

You should also update some routes to offer the possibility to encrypt the resulting PDFs:

* Chromium: https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/chromium/routes.go#L397
* PDF engines:https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/pdfengines/routes.go
* LibreOffice: https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/libreoffice/routes.go

vidschofelix:
regarding the routes: is it possible (or a good idea) to pass the multiPDFEngines to the chrome/libre routes? not sure atm how to access the encrypt parts from within the chrome/libre routes

gulien:
Both the Chromium and LibreOffice modules have a `PDFEngine`:

* https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/chromium/chromium.go#L66
* https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/libreoffice/libreoffice.go#L20

For the Chromium module, you should update the `convertURL` method (it is used by other methods): https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/chromium/routes.go#L327.

For the LibreOffice module, you should update the `convertRoute` method https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/libreoffice/routes.go#L16.

Also, it seems the following packages do not have all required tests (I suspect most of them being "stupid" test but still):

Capture d’écran 2022-09-14 à 14 02 43

pkg/modules/pdfengines/routes.go
vidschofelix: extracted into pkg/gotenberg/encryption.go. Hope thats ok

vidschofelix:
So i added encryption to every route that handles PDFs (all chromium, libreoffice and pdfengine). Restructured the openapi.yaml a little bit, so it's easier to maintain. Tested all routes locally. I hope i got all the tests, but it's hard to tell them all apart :D

vidschofelix:
any feedback?

gulien:
Hello @vidschofelix,

Sorry for the delay, but I don’t have enough time to properly review it.

There are many implementations available, so finding the correct interface for all modules is not straightforward.

Also, having an encrypt method will trigger a need for a `decrypt` route.

I keep this PR open until I find some time to work on it.

vidschofelix:
Ok, thank you :).

gomespauloho:
It's a very useful feature, I hope I can use it soon.

PR #547, opened 2022-11-15
Limit the number of concurrent chromium tasks

gstark:
When webhooks are used, it is difficult to manage the total number of requests being processed. Even when our system uses an event queue to submit requests, we'd have to await the webhook completion. In my case, the webhook posts to an AWS S3 URL, so the event queue would have to await the S3 hook.

This PR adds an option to limit the number of active instances that can exist simultaneously.

I'm not experienced in go, so the code may not be good quality. I also did not create any tests.


gulien:
Hello @gstark,

Your code looks good 👍 however, I'm not 100% convinced it should be merged in Gotenberg.

For instance, if I understand your use case correctly, you could also do the following:

1. Consume a message for a conversion.
2. Convert without the webhook feature (i.e., download the resulting PDF).
3. Upload on your side to S3.

gstark:
@gulien Thanks, I agree with the above and understand not merging.

CGvine:
Hello @gulien, actually I've been looking exactly for the solution that is implemented here. 
I want to limit the memory consumption of chromium, when using a https://gotenberg.dev/docs/modules/chromium#html request, since I always need to guarantee that a PDF is delivered, when requested.

Currently, the service could return 500 internal server errors, if too many parallel requests are incoming and the available memory exceeds. That already happens with e.g. a "<html><body>Hello world</body></html>" request, a memory limitation of 512MB (for the docker container) and 20 calls in parallel. Calling again in these cases is no good idea, I guess. 

Also, if the memory is exceeded, the caller usually gets a 503 service unavailable (if not getting a 500) and need to/could call again. 
It would be more convenient, if the request lasts as long as it needs and delivers the expected result, instead of a 503. 

These issues could be prevented by a max concurreny setting in many cases. (Finding a general solution for doing this automatically would be nice, but challenging, I guess)

Of course, if I missed something and the same thing can be achieved otherwise, I am glad for the information :)
Otherwise I would appreciate a merge.

gstark:
@CGvine, out-of-memory issues are part of what inspired this for me. I've implemented other features upstream to reduce the number of concurrent requests so that I avoid memory issues. My deployed instance is currently on Heroku so the memory limits are tight.

If you want to deploy this, you could take my branch and run `make build` to get a working docker image. If you need a published image, you could push this to `dockerhub` under your account. Your infrastructure could refer to this docker image.

If it helps, I've pushed this PR to a [dockerhub image under my account](gstark/gotenberg:chromium-limit-concurrent) - as always, be careful pulling images from untrusted/unverified sources.

gulien:
This PR is no longer relevant; there is now a single Chromium instance handling one request at a time (yet to be released).

Issue #550, opened 2022-11-24
Status code is not considered when calling the webhook


doriandekoning:
Currently it seems like the HTTP status code is not checked after a call to the webhook is made. Therefore if the server responds with a non 2xx status code this is not logged making it hard to monitor and debug. But I might be overlooking something.

The code where the response seems to be handled can be found in https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/webhook/client.go#L76 here only `err` as return value is checked but `resp.Status` is never used.

If you're open to it I'd happily open a merge request with a small change to ensure it returns an error if the status code is >=400 to ensure the calling function can log it.

gulien:
Hello @doriandekoning,

Good catch, feel free to open a PR. Just in case, could you also check the `client.do` method? It might also return an error on non `2xx` response.

doriandekoning:
Good one, as far as I know it doesn't but I'll double check to be sure!

Hopefully I have some time tomorrow or monday to write a patch :)

gulien:
Thanks @doriandekoning!

PR #551, opened 2022-11-25
fix(webhook/client): error on http status codes >= 400

doriandekoning:
As dicussed in https://github.com/gotenberg/gotenberg/issues/550 this change catches an http response code >=400 to ensure it the caller of the send function can handle it (in this case by logging it).

### DIFF
--- a/pkg/modules/webhook/client.go
+++ b/pkg/modules/webhook/client.go
@@ -3,6 +3,7 @@ package webhook
 import (
 	"fmt"
 	"io"
+	"net/http"
 	"strconv"
 	"time"
 
@@ -76,6 +77,10 @@ func (c client) send(body io.Reader, headers map[string]string, erroed bool) err
 	resp, err := c.client.Do(req)
 	if err != nil {
 		return fmt.Errorf("send '%s' request to '%s': %w", method, URL, err)
+	}
+
+	if resp.StatusCode >= http.StatusBadRequest {
+		return fmt.Errorf("send '%s' request to '%s': got status: '%s'", method, URL, resp.Status)
 	}
 
 	defer func() {
--- a/pkg/modules/webhook/middleware_test.go
+++ b/pkg/modules/webhook/middleware_test.go
@@ -339,6 +339,7 @@ func TestWebhookMiddlewareAsynchronousProcess(t *testing.T) {
 		expectWebhookFilename         string
 		expectWebhookErrorStatus      int
 		expectWebhookErrorMessage     string
+		returnedError                 *echo.HTTPError
 	}{
 		{
 			request: buildMultipartFormDataRequest(),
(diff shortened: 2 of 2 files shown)


gulien:
Thanks @doriandekoning!

FYI:

> retryablehttp performs automatic retries under certain conditions. Mainly, if an error is returned by the client (connection errors, etc.), or if a 500-range response code is received (except 501), then a retry is invoked after a wait period. Otherwise, the response is returned and left to the caller to interpret.

pkg/modules/webhook/client.go
```diff
@@ -78,6 +78,10 @@ func (c client) send(body io.Reader, headers map[string]string, erroed bool) err
 		return fmt.Errorf("send '%s' request to '%s': %w", method, URL, err)
 	}
 
+	if resp.StatusCode >= 400 {
```
gulien, line 81:
```suggestion
	if resp.StatusCode >= http.StatusBadRequest {
```

pkg/modules/webhook/client.go
```diff
@@ -78,6 +78,10 @@ func (c client) send(body io.Reader, headers map[string]string, erroed bool) err
...
+		return fmt.Errorf("send '%s' request to '%s' status: '%s'", method, URL, resp.Status)
```
gulien, line 82:
```suggestion
		return fmt.Errorf("send '%s' request to '%s': got status '%s'", method, URL, resp.Status)
```

doriandekoning:
Thanks for the review, I'll apply the suggestions. I didn't really see any unit tests for this file in `client_test.go` but I assume they live in a different place? 

Also if I see it correctly the [this](https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/webhook/client.go#L23) field is a struct not an interface so dependency injection (to have it return some non-ok status codes) would be a bit hard ;)  How do you normally test these kind of things?

gulien:
Best place to test it would be the `TestWebhookMiddlewareAsynchronousProcess` method from the `middleware_test.go` file. This method is a bit convoluted though 😅

doriandekoning:
Thanks! I'll add some cases there!

pkg/modules/webhook/middleware_test.go
```diff
@@ -387,6 +388,27 @@ func TestWebhookMiddlewareAsynchronousProcess(t *testing.T) {
...
+			}(),
+			mod: buildWebhookModule(),
+			next: func() echo.HandlerFunc {
+				return func(c echo.Context) error {
+					ctx := c.Get("context").(*api.Context)
+
+					return ctx.AddOutputPaths("/tests/test/testdata/api/sample3.pdf")
+				}
+			}(),
+			returnedError: echo.ErrInternalServerError,
```
gulien, line 406: > retryablehttp performs automatic retries under certain conditions. Mainly, if an error is returned by the client (connection errors, etc.), or if a 500-range response code is received (except 501), then a retry is invoked after a wait period. Otherwise, the response is returned and left to the caller to interpret.

Testing a `5xx` will trigger a retry, and fails with a standard error. You should test a `Bad Request` IMO.

doriandekoning: Ah right, I thought it then retries for n times and returns the final error, I've changed the test to a `Bad Request`.

PR #574, opened 2023-01-05
feat: support for libreoffice conversion to HTML

j-durbin:
## Optional HTML output format for libreoffice conversions

This PR adds an HTML output option to libreoffice conversions.

*Apologies in advance for any issues with the code, as I am not a golang'er.*

## Purpose?

Spreadsheets can render quite strangely in PDF, particularly when there are many columns.  For example, if the columns would exceed the width of the first page (and page range is not limited), all of the data corresponding to the columns that will fit will be in consecutive pages, then more columns with more pages, etc.  When the page range is limited, the columns are truncated.

The new functionality doesn't explicitly filter the extensions to limit HTML output for spreadsheets, since it could be useful for docs/etc.

## How to use?

Add the `htmlFormat=true` parameter to the libreoffice convert endpoint:
```
curl -XPOST localhost:3000/forms/libreoffice/convert --form 'files=@example.xlsx' --form 'htmlFormat=true' -o test.html
```

#### p.s. gotenberg is awesome, thanks!


gulien:
Hello @j-durbin!

Thanks for this PR! Alas, I really want to keep Gotenberg about PDFs, so I won't merge it.

That being said, you could create your own variant using a fork or even https://github.com/gotenberg/gotenberg-template-repository 😉 

Good luck!

PS: do not hesitate to open a PR on https://github.com/gotenberg/awesome-gotenberg if you want make your work "public"!

PR #578, opened 2023-01-09
Make prometheus metrics collect path customizable

edebernis:
This PR allows to customize path where Prometheus metrics are exposed. Currently, the path `/prometheus/metrics` is hardcoded.

Maybe, some additional constraints on this path could be discussed to avoid any conflicts with existing endpoints.

### DIFF
--- a/pkg/modules/prometheus/prometheus.go
+++ b/pkg/modules/prometheus/prometheus.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/gotenberg/gotenberg/v7/pkg/gotenberg"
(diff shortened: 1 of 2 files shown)


gulien:
Hello @edebernis,

What's your use case here? Note that you may also change the root path of all endpoints with `--api-root-path`.

edebernis:
We have a company standard to expose Prometheus metrics using `/metrics` route on every deployed services.

Obviously, we could do some URL rewriting or specific configuration on Prometheus collecting service but being able to configure it here would be probably better for maintainability.

gulien:
Ok, got it! It does make sense to have one `/metrics` route, like the `/health` route, but:

* Having the module name in the route pattern is actually a convention within Gotenberg (the only exception being the `/health` route, which gathers the health status of all modules).
* The Prometheus module is one implementation (and the only one for now :trollface:) for exposing metrics, but there could be another one in the future.

Allowing to change the route path goes against those "rules", but thanks for this well written PR though 👍 I feel a bit bad for closing it 😅

Issue #576, opened 2023-01-06
LibreOffice newer versions stability issues


gulien:
I did a bit of research and came to the following result with a simple binary search:

Working | Not Working | Release Date
-- | -- | --
  | 7.7.1 (4.4.2.3) | 03.12.2022
  | 7.6.1 (7.4.1.2 40(Build:2)) | 03.10.2022
7.6.0 (7.3.5.2 30(Build:2)) |   | 01.09.2022
7.5.4 (7.3.5.2) |   |  
7.5.3 (7.3.4.1) |   | 07.06.2022

The version in brackets represents the used LibreOffice version in each gotenberg version.
That might explain the sudden issue on our end, as we used `gotenberg/gotenberg:7` as the image. I've created this issue on the 04th of October, and the update with the broken version of LO was released on the 03rd of October. Our CI/CD pipeline most certainly did a `docker-compose pull` which updated to the named version and explains the sudden error(s).

So, basically - the culprit is on of 1.185 commits between v7.4.1.2 and v7.3.5.2 (https://github.com/LibreOffice/core/compare/libreoffice-7.4.1.2...libreoffice-7.3.5.2) - fun to debug tbh

_Originally posted by @PierreSchwang in https://github.com/gotenberg/gotenberg/issues/516#issuecomment-1372607904_

gulien:
Thanks @PierreSchwang for the investigation 👍

gulien:
https://github.com/gotenberg/gotenberg/issues/442#issuecomment-1384186104

Gotenberg garbage collector causing an issue with newer LibreOffice versions?

gulien:
https://github.com/gotenberg/gotenberg/issues/577#issuecomment-1401636345

yan-hic:
> [#577 (comment)](https://github.com/gotenberg/gotenberg/issues/577#issuecomment-1401636345)
> 
> Lock can take longer than 10 seconds?

Faced that exact same error re lock timeout right off the bat when starting a private `gotenberg:7`.

@gulien would you advise to downgrade 7.4.2 until fixed ? What major features would one miss ?
Our use case: converting xlsx in landscape pdf

gulien:
Hello @yan-hic,

You'll miss newer PDF Formats (PDF/A-2b & PDF/A-3b) but that's it 😉

yan-hic:
[...]

Issue #442, opened 2022-03-26
LibreOffice listener fails to start


gulien:
### Discussed in https://github.com/gotenberg/gotenberg/discussions/440

<div type='discussions-op-text'>

<sup>Originally posted by **Lincoln-hub** March 22, 2022</sup>
I am running Gotenbergr on my local computer, I installed all the required engines but I'm running into this error: [FATAL] starting uno: start long-running LibreOffice listener: execute LibreOffice listener: context done: context deadline exceeded exit status 1.  Does anyone know how to fix this?</div>

gulien:
@lincoln-hub could you disable the long-running LibreOffice listener with `--uno-listener-restart-threshold=0`? If that does not work, could you rollback to version `7.3`?

Lincoln-hub:
disabling the long-running LibreOffice listener didn't work either. version 7.3 gives the error below(looks like there is no option for --user-profile):
2022/03/26 22:17:55.067 DEBUG   api     form data values: map[] {"trace": "4ea59205-b1da-4385-b4cd-057e025ce49e"}
2022/03/26 22:17:55.067 DEBUG   api     form data files: map[Data_for_scaling.xlsx:/tmp/55c85e27-b816-4442-bf56-81e47da09a86/Data_for_scaling.xlsx]     {"trace": "4ea59205-b1da-4385-b4cd-057e025ce49e"}
2022/03/26 22:17:55.067 DEBUG   api     print to PDF with: {Landscape:false PageRanges: PDFArchive:false}       {"trace": "4ea59205-b1da-4385-b4cd-057e025ce49e"}
2022/03/26 22:17:55.067 DEBUG   api.cmd start unix process: /usr/bin/unoconv --user-profile ///tmp/bdd4323e-579f-445f-8c21-8a5c47cad9f8 --port 41383 --format pdf -vvv --output /tmp/55c85e27-b816-4442-bf56-81e47da09a86/26a72c11-882b-4ec2-9993-e82bb2d5d1a2.pdf /tmp/55c85e27-b816-4442-bf56-81e47da09a86/Data_for_scaling.xlsx   {"trace": "4ea59205-b1da-4385-b4cd-057e025ce49e"}
2022/03/26 22:17:55.169 DEBUG   api.cmd.stdout  unoconv: option --user-profile not recognized, try unoconv -h for a list of all the options     {"trace": "4ea59205-b1da-4385-b4cd-057e025ce49e"}
[...]

PR #589, opened 2023-01-13
Replace Debian with Ubuntu for greater control over LibreOffice versions

padpadthai:
Hi there,

I noticed the [issues](https://github.com/gotenberg/gotenberg/issues/576) caused by Debian's LibreOffice dependencies. By using Ubuntu Focal you get LibreOffice 7.3.7 which seems more stable than 7.4.x for PDF conversion.

In addition, you can also access more old LibreOffice versions from Ubuntu in the future via [Personal Package Archives](https://launchpad.net/~libreoffice/+archive/ubuntu/ppa) should regressions occur.

Let me know what you think.


gulien:
Thanks @padpadthai!

I did use Ubuntu previously, but, last time I've checked, recent Chromium versions were only available through the Snap package manager. I did not find a way to use this package manager in a Dockerfile, so I switched to Debian instead.

padpadthai:
Ok thanks for the info @gulien, I think I understand the nuance of the situation now.

Please have a look at the updated commit. It installs Chromium from Debian sources. I used the following sources as guidance for this approach:

- https://askubuntu.com/questions/1204571/how-to-install-chromium-without-snap/1206153#1206153
- https://askubuntu.com/questions/1179273/how-to-remove-snap-completely-without-losing-the-chromium-browser/1206502#1206502
- https://gist.github.com/ishad0w/7665cde882aa7dc3eec99613802e61e4

Would this work?

gulien:
I guess it would work, but it feels a bit hacky IMO 😄 also, @sharongut found out that the culprit could be the garbage collector from Gotenberg (see https://github.com/gotenberg/gotenberg/issues/442#issuecomment-1384186104). I'd prefer for now to look at this direction 😉

padpadthai:
Ok, I agree that the hacky-ness is not ideal here 😅. All the standard approaches I tried were blocked one way or another. Hopefully, you get this resolved via the garbage collector approach.

I will close this PR. Thanks for your feedback 😄

gulien:
Thanks for your time @padpadthai 😄

PR #603, opened 2023-02-12
added keepMergeOrder functionality

faizanazim11:
Added functionality to keep files order when converting from word to PDF and merging.

To enable this functionality added a new method for generating file output path that prefixes a counter with uuid for filename.

### DIFF
--- a/pkg/modules/api/context.go
+++ b/pkg/modules/api/context.go
@@ -204,6 +204,13 @@ func (ctx Context) GeneratePath(extension string) string {
 	return fmt.Sprintf("%s/%s%s", ctx.dirPath, uuid.New(), extension)
 }
 
+// GeneratePathOrdered generates a path within the context's working directory
+// and keeps the original files order. It does not create a file.
+
+func (ctx Context) GeneratePathOrdered(extension string, i int) string {
+	return fmt.Sprintf("%s/%d_%s%s", ctx.dirPath, i, uuid.New(), extension)
+}
+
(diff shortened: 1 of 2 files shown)


faizanazim11:
Hi  @gulien ,

In our use case we are sending some word files to gotenberg with merge as true, the functionality is working as expected but while merging it's order is changing. We experienced this in latest version code as well, when I checked the code I saw that you  are generating uuid for file names that you are converting into pdf and then after that you are reading those files from os and merging them. 
In this case as uuid are random string if the files stored are getting changed it will break the order of merge. To fix this issue I added a keepMergeOrder parameter which changes the method of generating file names by adding the file number as prefix to the uuid file name.

I might be wrong in doing this, can you confirm if this is expected or if you have any other solution/suggestion to my problem.

gulien:
Actually, the LibreOffice route does:

1. Get your files and sort them alphabetically. In other words, if you send `foo.docx` and `bar.docx`, the order will be first `bar.docx` and then `foo.docx`.
2. Then converts each file in PDF and create an array of PDF paths, keeping the previous order. Using our example, the order will be first `{uuid}.pdf` (i.e., `bar.docx`) and then `{uuuid}.pdf` (i.e., `foo.docx`).
3. Then merges using the order from the array of PDF paths.

So the most important part is to order your files alphabetically on the client side (see for instance https://github.com/gotenberg/gotenberg-php/blob/main/src/Modules/LibreOffice.php#L110).

PR #610, opened 2023-03-06
chore: add 503 to every endpoint as this is a known error

pieterocp:
As detailed on the website documentation, but not in the openapi spec.


gulien:
Thanks 👍

PR #649, opened 2023-05-23
Graceful shutdown on SIGINT and SIGTERM

kevin-lindsay-1:
A second SIGINT forces shutdown, as usual.

### DIFF
--- a/cmd/gotenberg.go
+++ b/cmd/gotenberg.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/signal"
+	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


kevin-lindsay-1:
@gulien I noticed that this container doesn't shutdown properly in K8S with graceful shutdown, because it immediately terminates upon receiving a SIGTERM, which is not the K8S contract.

In K8S, a SIGTERM is sent, graceful shutdown period is waited, and if it still hasn't exited, a SIGKILL is sent.

I made this edit via the GH UI, but it looks like it should work just fine.

If you'd like me to test a pre-release build or something, you can shoot me a container image/tag and I'll give it a whirl.

gulien:
@kevin-lindsay-1 seems to work as expected 👍

PR #676, opened 2023-08-22
run scheduled builds every week

ansnoussi:
- the current latest image for Gotenberg has several vulnerabilities (some from the base debian image, but mostly from installed packages like chromium)
- rebuilding the image will fetch latest versions for these packages and solve 17High, 17Medium and 7Low vulnerabilities:
- here's the test for rebuilding the image and scanning with Docker scout : 
  - before : 
Screenshot 2023-08-22 at 14 31 13
  - after : 
Screenshot 2023-08-22 at 14 31 37

## What's added in this PR : 
Run scheduled weekly builds to the keep the latest image always up to date.


gulien:
Hello @ansnoussi,

Thanks for your PR! AFAIK, it overrides the latest versions? Wouldn't it better to create a new minor release instead?

gulien:
Sorry, I meant a PATCH release 😬 I understand why it would make sense in some projects, but Gotenberg relies heavily on its dependencies. That’s why I tend to consider that the dependencies in the Docker image are part of the semantic versioning.

ansnoussi:
I understand your point. And ultimately, there could be a system with :
 - stable tag for the minor and major versions ( images tagged with `7`, `7.5` , `7.6`, ... will keep getting updated)
 - unique tag for minor versions ( images tagged with `7.5.1` , `7.5.2`, ... will not get any updates, but will be released on schedule)
I can make this change if you'd like, but it's important to have someway to keep getting security updates without using the `latest` tag which could introduce a breaking change.

having said that, I still think it's even better to rebuild all images on schedule, because I don't see a case where someone would prefer to use an older image version with known vulnerabilities rather than a newer image with the exact api, no breaking change, and fewer/no vulnerabilities at all.

PS: I do know that Gotenberg should not be public-facing, but this keeps triggering automated security alerts, which I believe should be the case for other users too.

gulien:
Thanks @ansnoussi and sorry for the delay.

I’m still not convinced of the actual implementation. When a bug comes, it’s often great to know the exact version that is affected. If we auto-update `7` version, the startup message will still show `7.9.1`.

IMO, I do think having a dedicated patch release is a better option in our context.

ansnoussi:
@gulien I do understand your hesitation, but I think the issue you are talking about can be fixed in 2 ways : 
- having automated tests done on the built image : best solution IMO but takes time/effort to put in place.
- fixing all dependencies versions down to minor release : that way you make sure no breaking change is introduced by any dependency : an overall best practice and should be done even if you implement the first solution.

From my POV, having regular updates is mandatory ti fix security findings, but releasing a patch version with each re-build of the image is not perfect : 
- if there were no updates to any dependency/base image => you would have many patch versions that are exactly the same, and it would be costly.
- I don't see a scenario where someone would rather use an old version with the **exact** gotenberg api, but vulnerable base image / dependencies.

I think to help you make a decision, you can take a look at other OS projects with nightly docker image updates : 
- https://github.com/Kong/kong/blob/master/.github/workflows/release.yml
- https://github.com/fedora-cloud/docker-brew-fedora/blob/master/.github/workflows/update_image.yml
- https://github.com/AdoptOpenJDK/openjdk-docker/blob/master/.github/workflows/updater.yml
- https://github.com/rocky-linux/sig-cloud-instance-images/blob/main/.github/workflows/build.yml
- https://github.com/adoptium/containers/blob/main/.github/workflows/updater.yml
- https://github.com/osrf/docker_images/blob/master/.github/workflows/trigger_nightly.yaml

PR #679, opened 2023-08-30
Address critical vuln in `debian:11-slim` 

adamstrawson:
Our security scanning tooling flagged that Gottenberg has a critical vulnerability because of it's use of `debian:11-slim`. This change bumps the base image to the latest non-vulnerable version.

The critical CVE is https://www.cve.org/CVERecord?id=CVE-2021-32292

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -30,7 +30,7 @@ RUN go build -o gotenberg -ldflags "-X 'github.com/gotenberg/gotenberg/v7/cmd.Ve
 # ----------------------------------------------
 # Final stage
 # ----------------------------------------------
-FROM debian:11-slim
+FROM debian:bookworm-20230725-slim
 
 ARG GOTENBERG_VERSION
 ARG GOTENBERG_USER_GID


gulien:
Hello, the latest base image uses a LibreOffice version that does not work in stateful mode.

adamstrawson:
Thanks @gulien - how should we proceed with this? Are we waiting for LibreOffice to work in stateful mode first, or are there other dependancies preventing an upgrade?

gulien:
I’ll first have to figure out why it’s not working as expected. I have a good lead on where to start, but it will require an amount of time I don’t have right now alas.

grrywlsn:
@gulien do you want to share the thoughts you have, we can investigate it in the meantime?

gulien:
I have a "module" that does garbage collection, i.e., deletes any files from the `/tmp` repository that are older than X seconds (X being > to the API request timeout). 

It worked great because every temporary files that was generated by the API, Chromium or LibreOffice had a lifecycle <= to the APIT request timeout.

But then, on popular demand, I introduced the "stateful" variant of LibreOffice. In earlier version, deleting folders like the user profile directory had 0 impact. But newer versions of LibreOffice do not like that. As they do work in stateless mode though, I do think that one of the problem is the abrupt deletion of files required by LibreOffice.

So, the idea is:

1. To remove the "global" garbage collector.
2. Separate each temporary files into distinct folders:
   a. One for the requests files and result files.
   b. One for Chromium.
   c. One for LibreOffice.
3. Have a dedicated garbage collector per needs (i.e., delete after X seconds A, but only on demand for B and C).

Will it be enough? Dunno, but that's my best guess for now.

PR #690, opened 2023-09-24
Automate Dynamic Installation of Custom Fonts in Gotenberg

lawfulsoftware:
This commit enables Gotenberg to automatically fetch, extract, and install custom fonts. A tarball URI, specified via an environment variable, serves as the source for these custom fonts.

The implementation offers two modes of font installation: 1) Ephemeral: Fonts are installed every time the container is deployed. 2) Persistent: Fonts are saved in a volume and updated only when a version number, specified via an environment variable, changes. This approach gives users granular control over font management and minimizes unnecessary downloads.

Additionally, a .gitattributes file has been added to enforce LF line endings for shell scripts, ensuring consistent behaviour across different operating systems.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Ensure LF line endings for shell scripts
+*.sh text eol=lf
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -56,7 +56,7 @@ RUN \
     # Install system dependencies required for the next instructions or debugging.
     # Note: tini is a helper for reaping zombie processes.
     apt-get update -qq &&\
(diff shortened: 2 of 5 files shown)


gulien:
Hello @lawfulsoftware!

Sorry for the delay. I really live the approach; however, I won't merge it into master as it will open the project to severe vulnerabilities AFAIK.

However, if you create you're own repository with this script + a documentation on how to use it, I'lll add a link in the documentation and the [awesome](https://github.com/gotenberg/awesome-gotenberg) repository.

gulien:
@lawfulsoftware 

Actually, I do not consider it to be a "severe" vulnerability, but, I'm confident some users will. And I don't want to deal with that 😄 

Also, there are valid alternatives to dynamic loading, like, building your own image, which make more sense for most scenarios.

PR #692, opened 2023-09-25
Added org.opencontainers.image.source label to allow dependabot pull release notes

mburumaxwell:
Dependabot uses the `org.opencontainers.image.source` label to fetch release notes, changelogs and commit history for docker images as per https://github.blog/changelog/2023-04-13-dependabot-now-supports-fetching-release-notes-and-changelogs-for-docker-images/

The label is also a standard name for opencontainers as per https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys


gulien:
Thanks @mburumaxwell!

PR #765, opened 2024-01-03
feat(chromium): capture screenshot

gulien:
## New Feature

One may now create full page screenshot with 3 new URLs:

* `/forms/chromium/screenshot/url`
* `/forms/chromium/screenshot/html`
* `/forms/chromium/screenshot/markdown`

These routes accept the following form fields:

* `format`: png, jepg, or wepb (default png).
* `quality` (jpeg only): 0 to 100 (default 100).
* `optimizeForSpeed` (bool): defines whether to optimize image encoding for speed, not for resulting size (default false).

The following features are also available:

* SkipNetworkIdleEvent
* FailOnHttpStatusCodes
* FailOnConsoleExceptions
* WaitDelay
* WaitWindowStatus
* WaitForExpression
* EmulatedMediaType
* OmitBackground


brusch:
@gulien very nice feature, thanks a lot! 👍 
I haven't seen the possibility to specify the window size for chrome, is that right? That would be a very nice improvement 😊 

Thanks again!

Issue #771, opened 2024-01-08
Keep original filenames in zip output


gulien:
### Discussed in https://github.com/gotenberg/gotenberg/discussions/769

<div type='discussions-op-text'>

<sup>Originally posted by **chrishughes20** January  8, 2024</sup>
Hello!

I am wondering if it's possible to somehow set a specific filename for PDF files that ultimately get zipped? I have cloned the project and am running with the log level set to debug. In the logs, I see the filename I want for each filename but then the output filename is changed to a random filename. Here is part of the log for one of my files:
`{"level":"debug","ts":1704735359.0625625,"logger":"api.formslibreofficeconvert.usrbinunoconv.stderr","msg":"Input file: /tmp/aa2cf6de-5a9c-47c6-8fe2-694c777b5798/visit-report_bart-simpson_2023-12-24.pdf","trace":"a1461428-daa0-47dc-adee-dace4c651afe"}
{"level":"debug","ts":1704735359.1313896,"logger":"api.formslibreofficeconvert.usrbinunoconv.stdout","msg":"Output file: file:///tmp/aa2cf6de-5a9c-47c6-8fe2-694c777b5798/f6d3268d-5a7d-4088-b65f-4eb87338a733.pdf","trace":"a1461428-daa0-47dc-adee-dace4c651afe"}`

The filename I would like to be used as the filename is `visit-report_bart-simpson_2023-12-24.pdf` rather than the output filename `f6d3268d-5a7d-4088-b65f-4eb87338a733.pdf`. 

Also, I'm a total Go newb and have been trying to play around with a cloned version of the project and have added a few `c.Logger().Debug("we are here")` and `fmt.Print("Inside of route")` inside of `pkg/modules/libreoffice/routes.go` but I don't see this output anywhere in the console or Docker logs after rebuilding and restarting the Docker container. Could this output be being saved to a log file or something somewhere or am I (likely) doing something wrong?

Thank you for your help!</div>

gulien:
@chrishughes20 actually there is a big caveat: if someone send two documents with the same filename but not the same extension, it does not work. Another solution would be to allow the user to send its own mapping, but meh.

chrishughes20:
[...]

PR #793, opened 2024-02-06
feat(libreoffice): preserve filenames for zipped files instead of using randomly generated UUID filenames

chrishughes20:
Addresses issue [771](https://github.com/gotenberg/gotenberg/issues/771)

**Why this feature?**

The filenames generated by the `libreoffice/convert` route returns randomly generated UUID filenames that are then archived in a zip file. For example, `document.docx` would then become `9790de13-5754-49db-aaee-51c8d599c0c0.pdf` after conversion.

**Solution**

Preserve the original filenames for conversion. The example file `document.docx` would then become `document.pdf`. This solution also address an edge case where there may be multiple filenames with the same filename but different extensions. `document.doc` and `document.docx` would then become `document.doc.pdf` and `document.docx.pdf`.

### DIFF
--- a/pkg/modules/api/context.go
+++ b/pkg/modules/api/context.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/google/uuid"
 	"io"
 	"mime/multipart"
 	"net/http"
@@ -13,7 +14,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/uuid"
 	"github.com/labstack/echo/v4"
 	"github.com/mholt/archiver/v3"
 	"go.uber.org/zap"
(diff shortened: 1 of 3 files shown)


gulien:
Hey @chrishughes20!

Are there some missing commits / changes? 

> This solution also address an edge case where there may be multiple filenames with the same filename but different extensions. document.doc and document.docx would then become document.doc.pdf and document.docx.pdf.

I wonder if, by default, we should only do `document.doc.pdf` and so on, because it would be easier to handle programmatically?

chrishughes20:
Hey @gulien !

Apologies, I did miss a commit/change. I have added my updated version of `libreoffice/routes.go`. I think keeping the default naming behavior is cleaner at only a small programmatic cost but am open to suggestions! If you think

> I wonder if, by default, we should only do document.doc.pdf and so on, because it would be easier to handle programmatically?

is a better approach I can make those changes.

gulien:
@chrishughes20 could you rebase from branch `feat/libreoffice-preserve-filenames-zip`? 😊

chrishughes20:
@gulien Absolutely! Done.

PR #803, opened 2024-02-16
feat(ExifTool): read/write metadata

gulien:


piyushsriv:
Hi @gulien, just checking with you. Is there anything that is stopping you from merging this? Can I help with anything?

gulien:
@piyushsriv I’m working on it!

piyushsriv:
Hi @gulien, 
I am not sure what tags you are trying to write but ExifTool allows only valid and writable tags to be overwritten. You can find some example tags [here](https://exiftool.org/TagNames/XMP.html#pdf).

For example - In my original [PR](https://github.com/gotenberg/gotenberg/pull/776) I was using below tags
```
curl --location 'localhost:3000/forms/libreoffice/convert' \
--form 'files=@"/tmp/test.docx"' \
--form 'metadata="{\"Creator\": \"Personio\", \"Producer\": \"Workzag\" }"' > libreoffice_test.pdf 
```

Can you try that? I just confirmed those tags are getting overwritten.

gulien:
Oh thanks I was trying to write random tags:

https://github.com/gotenberg/gotenberg/blob/a8238eb8a7d9b26aedf73b822120a1023b7c7d9c/pkg/modules/exiftool/exiftool_test.go#L178)

Is there a reason for only allowing "valid" tags?

gulien:
I also need to find a complete list for documentation purpose 🤔

piyushsriv:
> Is there a reason for only allowing "valid" tags?

One reason I can think of is security and another could be inflating file sizes by writing millions or billions of random tags, maybe.

gulien:
@piyushsriv Long overdue, but `gotenberg/gotenberg:edge` should be available ~1 hour after this comment.

You may now write metadata using the form field `metadata` for the following routes:

* `/forms/chromium/convert`
* `/forms/pdfengines/merge`

The `metadata` accepts, in JSON format, the following value types: boolean, int, float, string, strings.

There are also two new routes:

* `/forms/pdfengines/metadata/read` - returns the metadata of the given PDFs in a JSON response (filename => metadata)
* `/forms/pdfengines/metadata/write` - write given metadata in one or more PDFs

piyushsriv:
Hi @gulien, 
Thanks for all the support.

Issue #463, opened 2022-05-30
Feature request: Limit the queue size for stateful mode LibreOffice conversions


NathanLamSeekasia:
The stateful mode LibreOffice conversion is nice because it is faster than stateless mode.
I tried some performance tests by launching multiple instances (stateful mode) with a load balancer in the front.

### Problem
The conversion queue size in Gotenberg looks unlimited.
So Gotenberg will keep accepting a new request no matter how many pending conversion is. 

End up the instances may:
- Be easily overloaded if there is a spike or unbalanced traffic
- Most of the requests will keep being timeout (loop in client side: timeout->retry->timeout->retry...)
- The overloaded instances cannot be recovered even scale up more instances after a while
- A possibility that Gotenberg started a conversion but the API request just timed out at the same time (wasted conversion)

### Suggestion
It would be good if having an option to limit the queue size of LibreOffice stateful mode.
And it will **throttle/reject the new request immediately** (avoid too many requests waiting for timeout) when there are a lot of pending conversions (based on `libreoffice_listener_queue_length`).

Then, client side can monitor easily and based on the rejected number to scale up more instances

Thanks and appreciated! 🙏

gulien:
Hello @NathanLamSeekasia,

Thanks for the well written issue!

It make sense, alas, I don't have much time to implement a solution right now. PR welcome though 😄

timgrohmann:
I think I can look into this and create a PR. We're also running into the issue of "old" Gotenberg instances only responding with timeouts, even if no documents are sent for a long time.

If I understand the code correctly, the queue is not specific to the LibreOffice module, so adding a parameter to set a maximum queue size would influence all modules.

Would that be acceptable?

gulien:
Hey @timgrohmann 👋

Both the Chromium and LibreOffice modules have their own queue length, although they are both using the supervisor struct. I guess one could implement a solution in the [Run](https://github.com/gotenberg/gotenberg/blob/975a9f53444173a695eac7640b9fba7cf5fa46e2/pkg/gotenberg/supervisor.go#L169) method, for instance, returning an error immediately if the queue size is above a threshold.

PR #804, opened 2024-02-17
feat: enable max queue size for chromium and libreoffice

timgrohmann:
See #463

### DIFF
--- a/pkg/gotenberg/supervisor.go
+++ b/pkg/gotenberg/supervisor.go
@@ -13,6 +13,10 @@ import (
 // to restart an already restarting [Process].
 var ErrProcessAlreadyRestarting = errors.New("process already restarting")
 
+// ErrMaximumQueueSizeExceeded happens if Run() is called but the maximum queue
+// size is already used.
+var ErrMaximumQueueSizeExceeded = errors.New("maximum queue size exceeded")
+
 // Process is an interface that represents an abstract process
 // and provides methods for starting, stopping, and checking the health of the
 // process.
@@ -74,6 +78,7 @@ type processSupervisor struct {
 	logger          *zap.Logger
 	process         Process
 	maxReqLimit     int64
+	maxQueueSize    int64
 	mutexChan       chan struct{}
 	firstStart      atomic.Bool
 	reqCounter      atomic.Int64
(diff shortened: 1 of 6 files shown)


gulien:
Thanks @timgrohmann! Great PR!

I'll add the missing tests.

gulien:
Closes #463.

pkg/gotenberg/supervisor.go
```diff
@@ -167,6 +173,11 @@ func (s *processSupervisor) Healthy() bool {
 }
 
 func (s *processSupervisor) Run(ctx context.Context, logger *zap.Logger, task func() error) error {
+	currentQueueSize := s.reqQueueSize.Load()
+	if s.maxQueueSize > 0 && currentQueueSize >= s.maxQueueSize {
```
b-viguier, line 177: Sorry to wake up this old PR, but I'm wondering if there isn't a potential concurrency issue here.

I'm not a Go expert, but AFAIK atomics only guarantee safe read/write access to a variable, but this is not comparable to a _critical section_ created with a mutex. Here, it means that `10` Goroutines could read the `reqQueueSize` value at the _same_ time, succeed to compare it to `maxQueueSize=5` and then incrementing all together the counter. The increment is safe (because it uses atomic), but it seems to be a gap in the logic because we will end up with a queue with `10` elements, isn't it?

We should allow other Goroutines to read the value only once comparison and potential increment are done IMHO 🤔 
Am I missing something?
Thx for your feedback 🙇

gulien: Hey thanks for the feedback @b-viguier!

You’re actually right! Did you experience some issues on your side? Also, do you want to try to fix it in a PR with a _mutex_ instead?

b-viguier: > Did you experience some issues on your side?

Nope, I tried to send _a lot_ of concurrent requests locally, and never detected any _overflow_... So I guess this is really an edge case

> Also, do you want to try to fix it in a PR with a mutex instead?

Unfortunately, I'm already running out of time, and Go is not my usual dev stack... But if you think it may help, I can open an issue with a clear description of the problem, in case someone else want to give it a try?

BTW: thanks for your great work on this tool 👏🙂

gulien: No problem! 😉 But if you find time to create an issue, it would be great to help me tracking this edge case! Thanks 😊

PR #843, opened 2024-03-22
feat(libreoffice): add exportFormFields option

mrbech:
Feel free to just close if not interested in adding this feature.

Exposes libreoffice's ExportFormFields option, defaults to `true` as this is the current behavior.
When set to `false` the output PDF will, instead of having form fields, just have the inputted/selected content of the fields.

For context I have some docx documents with selection fields that becomes pretty ugly in the output PDFs when exported as PDF form fields.

### DIFF
--- a/pkg/modules/libreoffice/api/api.go
+++ b/pkg/modules/libreoffice/api/api.go
@@ -50,6 +50,11 @@ type Options struct {
 	// Optional.
 	PageRanges string
 
+	// Form fields from input is exported as form fields in the resulting PDF.
+	// Defaults to true
+	// Optional.
+	ExportFormFields bool
+
 	// PdfFormats allows to convert the resulting PDF to PDF/A-1b, PDF/A-2b,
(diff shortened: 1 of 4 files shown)


pkg/modules/libreoffice/api/api.go
```diff
@@ -50,6 +50,11 @@ type Options struct {
 	// Optional.
 	PageRanges string
 
+	// Form fields from input is exported as form fields in the resulting PDF.
+	// Defaults to true
+	// Optional.
```
gulien, line 55:
```suggestion
	// ExportFormFields allows to... export form fields in the resulting PDF.
	// Optional.
```

gulien:
Looks great! A small suggestion regarding the godoc, otherwise 👍 thanks a lot!

gulien:
@mrbech `gotenberg/gotenberg:edge` is available with your update!

mrbech:
Perfect @gulien, thanks!

PR #844, opened 2024-03-25
feat(libreoffice): add singlePageSheets option

mrbech:
Exposes libreoffice's SinglePageSheets option, defaults to `false` as this is the current behavior. 

Normally libreoffice will take each sheet in a spreadsheet and fit into onto 1 or more A4 pages in the output PDF.
When `SinglePageSheets` is set to `true` libreoffice will instead render the entire spreadsheet as a single page, with the size of the page being whatever fits the content.

This is useful when the output PDF is to be used digitally and not printed to paper.

### DIFF
--- a/pkg/modules/libreoffice/api/api.go
+++ b/pkg/modules/libreoffice/api/api.go
@@ -54,6 +54,10 @@ type Options struct {
 	// Optional.
 	ExportFormFields bool
 
+	// SinglePageSheets allows to output each sheet as a single page in the resulting PDF.
+	// Optional
+	SinglePageSheets bool
+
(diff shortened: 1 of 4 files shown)


gulien:
Hey @mrbech,

Thanks for this new PR! A stupid question, but I guess it concerns only XLSX etc.?

mrbech:
> Hey @mrbech,
> 
> Thanks for this new PR! A stupid question, but I guess it concerns only XLSX etc.?

Should be, from my testing it works for xlsx and makes no difference when converting docx documents at least. 
The documentation I have been able to find isn't explicit about which formats this works for, but I would guess any formats that normally gets opened by libre office calc.

gulien:
Thanks again!

PR #848, opened 2024-04-03
feat(chromium): set cookies

### LINKED ISSUE #658: Possibility to send cookies along with Chromium requests
I am trying to render a SPA (single-page application) output to PDF using Chromium in Gotenberg. This SPA performs an API call, which requires a cookie for authentication. It seems to be impossible to set this cookie in Chromium using the [Gotenberg Chromium API](https://gotenberg.dev/docs/modules/chromium).

I have tried setting the Cookie using the `extraHttpHeaders` field, but that didn't work - I expect it sets the Cookie only on the initial (SPA/frontend) request, and not in any subsequent backend API calls.

Could we add an additional `cookies` form field for Chromium requests, which will be added to the cookie jar in Chromium so it automatically sends them for every (matching) request, including Javascript-initiated API calls?

gulien:
Closes #658.

Looking for testers - `gulnap/gotenberg:8.3.0-set-cookies`.

New form field `cookies` available within the Chromium module. It accepts a JSON-formatted array with items accepting the following keys:

* `name` (required)
* `value` (required)
* `domain` (required)
* `path`
* `secure` - boolean
* `httpOnly` - boolean
* `sameSite` - accepted values, beside empty, are: "Strict", "Lax" or "None"

Cookies are set to expire after the conversion.


Sungq1990:
When will this 'feat' branch be merged into the 'main' branch and pushed to Docker Hub?

gulien:
@Sungq1990 I first need people to test this feature using the `gulnap/gotenberg:8.3.0-set-cookies` image.

Sungq1990:
@gulien I tested it.My page requires login, otherwise it will redirect to the login page. After setting the cookie, I successfully reached the target page after login.
image
image

gulien:
Thanks a lot @Sungq1990!

PR #866, opened 2024-04-26
Write log time in local timezone rather than UTC

jaime-pineiro-imatia:
Time in logs was printed in utc timezone, with this change we can set environment variable TZ=XXX and time is printed in that timezone.


gulien:
Hello @jaime-pineiro-imatia!

Thanks for your PR! What happens if there is no TZ variable? Is there one by default?

jaime-pineiro-imatia:
I've tested in my local environment, if i don't set TZ it prints time in UTC, as with "ts.UTC()"

gulien:
It seems indeed to be the case!

> // Local represents the system's local time zone.
// On Unix systems, Local consults the TZ environment
// variable to find the time zone to use. No TZ means
// use the system default /etc/localtime.
// TZ="" means use UTC.
// TZ="foo" means use file foo in the system timezone directory.
var Local *Location = &localLoc

Thanks! `gotenberg/gotenberg:edge` should be updated within an hour 👍

PR #901, opened 2024-06-12
feat(chromium): add user agent override

### LINKED ISSUE #900: `User-Agent`, `Accept-Encoding` and `Host` headers
I tried to generate a PDF using a Gotenberg container, but the resulting file was completely blank. I sent a POST request with [this URL](https://www.sec.gov/Archives/edgar/data/1067983/000119312524133049/d759664d8k.htm), but all I got was a white, empty page. You need to set the HTTP headers like [this](https://www.sec.gov/os/accessing-edgar-data). However, when I manually downloaded the HTML file and sent it to my Gotenberg container, the conversion worked like a charm. This made me think that maybe the `extraHttpHeaders` parameter isn't working as it should for `User-Agent`, `Accept-Encoding` and `Host`.

To get to the bottom of this, I was wondering if there's a way to take a closer look at the actual headers and body that a Gotenberg instance sends when it makes a GET request.

_Originally posted by @wjkoh in https://github.com/gotenberg/gotenberg/discussions/842#discussioncomment-9752869_

gulien:
Closes #900.

New form field `userAgent`.


gulien:
@wjkoh I have tested it successfully!

The image `gotenberg/gotenberg:edge` should be available within ~1 hour. I plan to release a new version once https://github.com/gotenberg/gotenberg/pull/898 is merged, but the `edge` tag is pretty stable.

wjkoh:
Wow, it works like magic! Thanks @gulien for such a quick fix! This update allows us to remove a lot of unnecessary code from our app. 😆 Once again, thank you so much, Julien!

Issue #915, opened 2024-07-01
Restart LibreOffice right away


MaxLap:
Once `--libreoffice-restart-after` is reached. The next call to Gotenberg will be much slower due to the restarting happening.

This would be a lot less noticeable if the restarted was done right after the `--libreoffice-restart-after`-nth request.

Right now, if I have a request every 10 seconds. Then they take about ~200ms each, except once every 10 request, the request takes ~2000ms. With a change as proposed, The ~2000ms would never be seen and it would always take only ~200ms.

Personally, I don't see a reason for this to be a configuration and should just be the default. It seems beneficial in basically all cases. Thoughts?

I'd be willing to look into making a PR for this if you are interested.

gulien:
Hello @MaxLap,

Indeed, that would be better. I did try to implement such a solution, but failed to build something stable enough. Most of the magic happens here: https://github.com/gotenberg/gotenberg/blob/main/pkg/gotenberg/supervisor.go.
That being said, under heavy load, restarting before or after is the same 😄

PR #916, opened 2024-07-01
Eagerly auto restart processes

MaxLap:
Before: The auto-restart was done at the beginning of a request, creating visible latency every time.

This change makes it so that processes are restarted after a request instead. When not under full load, this can reduce / remove the latency.
Discussed in https://github.com/gotenberg/gotenberg/issues/915.

Closes https://github.com/gotenberg/gotenberg/issues/915


gulien:
@MaxLap Thanks a lot for your PR! Run `make fmt` to fix the lint issue 👍

MaxLap:
How to say you're new to Go without saying it xD Done!

It's weird that `go fmt` says "no Go files in /home/max/projects/gotenberg"...

Anyways, managed to install gofumpt and run the requested command.

pkg/gotenberg/supervisor_test.go
```diff
@@ -451,6 +451,10 @@ func TestProcessSupervisor_Run(t *testing.T) {
 				return
 			}
 
+			// Making sure restarts are finished
```
gulien, line 454:
```suggestion
			// Making sure restarts are finished.
```

gulien (changes requested):
The solution was quite simple actually lol

pkg/gotenberg/supervisor.go
```diff
@@ -212,18 +215,26 @@ func (s *processSupervisor) Run(ctx context.Context, logger *zap.Logger, task fu
...
 				if s.maxReqLimit > 0 && s.reqCounter.Load() >= s.maxReqLimit {
-					s.logger.Debug("max request limit reached, restarting...")
-					err := s.runWithDeadline(ctx, func() error {
-						return s.restart()
-					})
-					if err != nil {
-						return fmt.Errorf("process restart before task: %w", err)
-					}
+					s.logger.Debug("max request limit reached, restarting eagerly...")
+					// Will be done manually after the restart
```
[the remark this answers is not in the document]
MaxLap: I expanded on it:
```
// mutexChan will be released manually (not as part of the defer) after the restart,
// because restart needs the lock but is happening in the background.
```

pkg/gotenberg/supervisor.go
```diff
@@ -212,18 +215,29 @@ func (s *processSupervisor) Run(ctx context.Context, logger *zap.Logger, task fu
...
-					err := s.runWithDeadline(ctx, func() error {
-						return s.restart()
-					})
-					if err != nil {
-						return fmt.Errorf("process restart before task: %w", err)
-					}
+					s.logger.Debug("max request limit reached, restarting eagerly...")
+
+					// mutexChan will be released manually (not as part of the defer) after the restart,
+					// because restart needs the lock but is happening in the background.
```
gulien, line 224: It's actually confusing more than anything lol

Better to remove this comment altogether, as the code base makes sense without it anyway! 😉

gulien (changes requested):

gulien:
Almost done, thanks!

gulien (approved):

pkg/gotenberg/supervisor.go
```diff
@@ -212,18 +215,26 @@ func (s *processSupervisor) Run(ctx context.Context, logger *zap.Logger, task fu
...
-						return s.restart()
-					})
-					if err != nil {
-						return fmt.Errorf("process restart before task: %w", err)
-					}
+					s.logger.Debug("max request limit reached, restarting eagerly...")
+					releaseMutexChan = false
+
+					go func() {
+						err := s.runWithDeadline(ctx, func() error {
```
gulien, line 225:
```suggestion
						err := s.runWithDeadline(context.Background(), func() error {
```

We should not use the request context here, as it may be cancelled before the process restart. Using `context.Background` is pretty safe, as all processes have a start timeout anyway.

gulien (changes requested):

gulien (approved):

gulien:
Thanks a lot for your time @MaxLap!

gulien:
@MaxLap `gotenberg/gotenberg:edge` should be updated within ~1 hour.

MaxLap:
Seems to work great on my side. Thank you!

PR #957, opened 2024-09-05
fix for go: download go1.23 toolchain not available

gabriel-vasile:
The problem appears when trying to run locally:
```bash
➜  gotenberg git:(main) ✗ go run cmd/gotenberg/main.go
go: downloading go1.23 (darwin/arm64)
go: download go1.23 for darwin/arm64: toolchain not available
```

The change is based on this answer: https://github.com/golang/go/issues/62278#issuecomment-1693538776

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/gotenberg/gotenberg/v8
 
-go 1.23
+go 1.23.0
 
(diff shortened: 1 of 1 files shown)


gulien:
TIL 😄 thanks for your PR!

PR #965, opened 2024-09-08
feat: optimize PDF

### LINKED ISSUE #359: Feature Request: Add a way to compress final pdf size
The currently generated pdfs look great but could be smaller in size.

If there would be an option/setting to choose the quality (1-100) for the pdf output that would be great.

gulien:
Closes #359.

A mixte of QPDF (compressing streams) and LibreOffice (images). Not perfect, but good enough until a better tool is available.


gulien:
For now, the best option that does not break too much stuff (e.g., PDF compliance, metadata, etc.) is using the `/forms/libreoffice/convert` route. I've updated the documentation to reflect this.

Until a better tool is found.

PR #971, opened 2024-09-14
feat(api): add 'downloadFrom' form field

### LINKED ISSUE #894: Feature request: office document to PDF via URL
My team would like a new route, similar to `POST /forms/libreoffice/convert`, but instead of accepting files as a multipart upload, the endpoint accepts a list of `fileUrls`.

The route would download each file, and then call the existing libreoffice conversion.

The benefits of this are:
1. Removes upload size restrictions. For example #540 is not practical today b/c Lambda has a 6MB body limit, so converting office files to PDF would only work for small files
2. Improves performance for anyone with remote file storage. For example, my team uses S3. To use Gotenberg, we first have to download a file from S3, then upload it to Gotenberg. It'd be faster if Gotenberg did the S3 download directly

gulien:
Closes #894.


pkg/modules/api/context.go
```diff
@@ -126,6 +138,144 @@ func newContext(echoCtx echo.Context, logger *zap.Logger, fs *gotenberg.FileSyst
...
+
+				contentDisposition := resp.Header.Get("Content-Disposition")
+				if contentDisposition == "" {
+					return WrapError(
+						fmt.Errorf("no 'Content-Disposition' header from '%s'", dl.Url),
+						NewSentinelHttpError(http.StatusBadRequest, fmt.Sprintf("No 'Content-Disposition' header from '%s'", dl.Url)),
+					)
+				}
+
+				_, params, err := mime.ParseMediaType(contentDisposition)
```
aldenquimby, line 229: @gulien this line is causing my application to fail with the error `mime: no media type`. My current file server returns the following:
```
Content-Disposition: ";filename*=UTF-8''tips-building-social-media-presence.pdf"
```

According to [rfc6266](https://www.rfc-editor.org/rfc/rfc6266#page-5), that's valid and the lack of disposition type should be treated as "attachment"

But the go ParseMediaType fails to parse that... any of these work instead:
```
Content-Disposition: "attachment;filename*=UTF-8''tips-building-social-media-presence.pdf"
Content-Disposition: "inline;filename*=UTF-8''tips-building-social-media-presence.pdf"
Content-Disposition: "#;filename*=UTF-8''tips-building-social-media-presence.pdf"
Content-Disposition: "anything_but_whitespace_here;filename*=UTF-8''tips-building-social-media-presence.pdf"
```

I will see if I can change my file server response headers... 

Do you have any experience contributing to `go`? I have never used it before. I'm wondering if instead of change my file server, I can [submit a patch here](https://github.com/golang/go/blob/165bf241f2f7c72cffd83e278d674ae3ddbd72a1/src/mime/mediatype.go#L105-L107)

aldenquimby: I logged https://github.com/golang/go/issues/69551 - if that seems reasonable to you and you're willing to give it a 👍 , I'd appreciate that

gulien: I gave it a 👍 

Thanks for the investigation. I did not expect Golang to fail there 🫤 

In the meantime, it may be possible to force the "attachment"? For instance: https://sweemeng.medium.com/setting-content-disposition-in-s3-9598cefc2ea7

aldenquimby: Update - I confirmed that if I change my content-disposition header coming from the file server, this new code works for AWS S3 👍

PR #1099, opened 2025-01-15
feat(api): added dummy root route pointing to docs

nktnet1:
Added the root route, `/`:

feat-root-route

similar to what is present at https://demo.gotenberg.dev:

demo-root-route

---

Only a minor issue, but users can get confused if nothing is displayed at the root URL, e.g. in [this comment](https://github.com/coollabsio/coolify/pull/4759#issuecomment-2589839139).

This could occur when someone is trialing the software using a "one-click deploy" feature/template, e.g. offered by Coolify/Dokploy.
Also potentially useful for basic/automatic health-check services that may not use the `/heath` route.

I believe it's not uncommon to avoid having `/` as 404 - e.g. here's what conduit displays at `/`:

conduit

---

### DIFF
--- a/pkg/modules/api/api.go
+++ b/pkg/modules/api/api.go
@@ -483,18 +483,20 @@ func (a *Api) Start() error {
 		)
 	}
 
-	// Root routes.
+	// Root route.
 	a.srv.GET(
-	    a.rootPath, 
-	    func(c echo.Context) error {
-		c.HTML(http.StatusOK, `Hey, this Gotenberg no UI, it's an API. Head to the <a href="https://gotenberg.dev">documentation</a> to learn how to interact with it 🚀`)
-	    },
+		a.rootPath, 
+		func(c echo.Context) error {
+				return c.HTML(http.StatusOK, `Hey, this Gotenberg has no UI, it's an API. Head to the <a href="https://gotenberg.dev">documentation</a> to learn how to interact with it 🚀`)
+		},
 	)
+
+	// Favicon route.
 	a.srv.GET(
-	    fmt.Sprintf("%sfavicon.ico", a.rootPath) 
-	    func(c echo.Context) error {
-		return c.NoContent(http.StatusNoContent)
-	    },
+		fmt.Sprintf("%sfavicon.ico", a.rootPath),
+		func(c echo.Context) error {
+			return c.NoContent(http.StatusNoContent)
+		},
 	)
 
 	// Let's not forget the health check routes...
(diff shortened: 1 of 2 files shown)


pkg/modules/api/api.go
```diff
@@ -483,6 +483,18 @@ func (a *Api) Start() error {
...
+	a.srv.GET(a.rootPath, func(c echo.Context) error {
+		html := `
+		<p>
+			Hey, Gotenberg has no UI, it's an API. Head to the
+			<a href="https://gotenberg.dev" target="_blank">documentation</a>
+			to learn how to interact with it 🚀
+		</p>
+	`
+		return c.HTML(http.StatusOK, html)
+	})
```
gulien, line 496:
```suggestion
	// Root routes.
	a.srv.GET(
	    a.rootPath, 
	    func(c echo.Context) error {
		c.HTML(http.StatusOK, `Hey, this Gotenberg no UI, it's an API. Head to the <a href="https://gotenberg.dev">documentation</a> to learn how to interact with it 🚀`)
	    },
	)
	a.srv.GET(
	    fmt.Sprintf("%sfavicon.ico", a.rootPath) 
	    func(c echo.Context) error {
		return c.NoContent(http.StatusNoContent)
	    },
	)
```

pkg/modules/api/api_test.go
```diff
@@ -866,9 +866,16 @@ func TestApi_Start(t *testing.T) {
 				return
 			}
 
-			// health requests.
 			recorder := httptest.NewRecorder()
 
+			// root request.
+			rootRequest := httptest.NewRequest(http.MethodGet, "/", nil)
```
gulien, line 872: Add favicon.ico "test" 😄

nktnet1:
Hey @gulien, I haven't programmed in Go before so I'm not sure why the test was failing for `/favicon.ico`.

The test output says `200` was given for `/favicon.ico` when 204 was expected, so my guess is that the `recorder` variable was being re-used from the root request (i.e. it can only record one response?)

I added the reassignment of recorder to a new `httptest.NewRecorder()` for each check/case, and the test appears to be passing. Not sure if this is the correct way though, as I saw the `recorder` variable in the original code being shared between `/health` and `/version` (unless that's a mistake, and it was missed because they're both status code 200?).

Thanks ~

gulien:
@nktnet1 I'll take a look 👍

gulien:
Thanks @nktnet1!

PR #1105, opened 2025-01-22
Add support for flattening annotations

chakalov:
This pull request introduces a new feature to Gotenberg that adds support for flattening annotations in PDF documents. Annotation or form flattening refers to the operation that changes annotations (such as markup, widgets, 3D models, etc.) into a static area that is part of the PDF document, just like the other text and images in the document.


chakalov:
Tested manually with:
```sh
curl \
--request POST http://localhost:3000/forms/pdfengines/flatten \
--form files=@/Users/chakalov/workspace/gotenberg/test/testdata/pdfengines/sample3.pdf \
-o my.pdf
```

gulien:
Hey @chakalov! 

Thanks a lot for this well written PR!

I wonder to what extend would it be possible to add it at the end of a process conversion? For instance, you may add [metadata](https://gotenberg.dev/docs/routes#metadata-chromium) using a PDF engine.

Would it be possible to as such a behavior to the following routes?

- `/forms/pdfengines/merge`
- `/forms/libreoffice/convert`
- `/forms/chromium/convert/url`
- `/forms/chromium/convert/html`
- `/forms/chromium/convert/markdown`

chakalov:
> Hey @chakalov!
> 
> Thanks a lot for this well written PR!
> 
> I wonder to what extend would it be possible to add it at the end of a conversion process? For instance, you may add [metadata](https://gotenberg.dev/docs/routes#metadata-chromium) using a PDF engine.
> 
> Would it be possible to add such a behavior to the following routes?
> 
> * `/forms/pdfengines/merge`
> * `/forms/libreoffice/convert`
> * `/forms/chromium/convert/url`
> * `/forms/chromium/convert/html`
> * `/forms/chromium/convert/markdown`

Hey @gulien!

Thank you for your feedback and the kind words about my PR!

I have made the requested changes and added the flatten behavior to the following routes:
* `/forms/pdfengines/merge`
* `/forms/chromium/convert/url`
* `/forms/chromium/convert/html`
* `/forms/chromium/convert/markdown`

However, I have some concerns regarding the flatten behavior for the `/forms/libreoffice/convert` and `/forms/chromium/convert/*` routes. I'm not entirely sure if the Chromium and LibreOffice conversion processes add any annotations that would require the use of the flatten option. Let me know what you think!

gulien:
Oh, indeed, it might be unnecessary with Chromium, as it generates a static PDF. However, I think LibreOffice is generating annotations, for instance when exporting notes.

chakalov:
No problem, I reverted it

chakalov:
@gulien the lint issue is a false positive (comparing struct to interface). It's described in more details here - https://github.com/golangci/golangci-lint/issues/1372

gulien:
@chakalov Some missing coverage (yeah, not great, see #1107), otherwise 👍

pkg/modules/qpdf/qpdf.go
```diff
@@ -101,6 +101,27 @@ func (engine *QPdf) Merge(ctx context.Context, logger *zap.Logger, inputPaths []
 	return fmt.Errorf("merge PDFs with QPDF: %w", err)
 }
 
+// Flatten merges annotation appearances with page content, deleting the original annotations.
+func (engine *QPdf) Flatten(ctx context.Context, logger *zap.Logger, inputPath, outputPath string) error {
+	var args []string
+	args = append(args, "--generate-appearances")
+	args = append(args, "--flatten-annotations=all")
+	args = append(args, inputPath)
+	args = append(args, outputPath)
```
gulien, line 110: I wonder if it would make more sense to use the flag [--replace-input](https://qpdf.readthedocs.io/en/stable/cli.html#option-replace-input) instead of creating a new file 🤔 

```suggestion
func (engine *QPdf) Flatten(ctx context.Context, logger *zap.Logger, inputPath string) error {
	var args []string
	args = append(args, "--generate-appearances")
	args = append(args, "--flatten-annotations=all")
	args = append(args, "--replace-input")
	args = append(args, inputPath)
```

chakalov: makes sense, will do

gulien (approved):

gulien:
Looks great! Thanks @chakalov! `gotenberg/gotenberg:edge` should be available in 15 minutes with this update 😉

PR #1108, opened 2025-01-23
ci(workflows): improvements

gulien:


.github/workflows/__delete.yml
```diff
@@ -0,0 +1,24 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  delete:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Delete
+        uses: gotenberg/gotenberg-release-action/actions/delete@main
```
bendavies, line 19: @gulien is the repo with this action not public?

bendavies: nevermind,i see they are local now

gulien: @bendavies hat brought you here? 😄

Issue #1101, opened 2025-01-16
Docker image violates MIT license


pschumacher:
While Gotenberg itself is published under the MIT license, the Docker image contains MS Core Fonts (ttf-mscorefonts-installer_3.8.1_all.deb). 

The font's EULA reads as follows:

> Reproduction and Distribution. You may reproduce and distribute an unlimited number of copies of the SOFTWARE PRODUCT; provided that each copy shall be a true and complete copy, including all copyright and trademark notices, and shall be accompanied by a copy of this EULA. Copies of the SOFTWARE PRODUCT may not be distributed for profit either on a standalone basis or included as part of your own product.

This conflicts with the MIT license and as a result, it is not possible to use the Docker image for commercial products.

It would be nice to have an image that complies with the MIT license. Perhaps the fonts could be removed from the image and added optionally during container runtime.

gulien:
@pschumacher huh, I never checked the EULA of this font, good catch!

However, I'm kinda at a crossroad here:

- These fonts are sort of mandatory, as Microsoft fonts are used in a lot of Office documents.
- Shipping a *8.x* version without these fonts cannot be done without breaking workflows in userland.
- These fonts conflict with the MIT license 🤔 

I *could* ship a new *9.x* version though without this fonts, but I feel like it's too early for that.

gulien:
@andrioid you mean, it fetches them at runtime?

The-MAZZTer:
> Some open-source projects, like Wine solve this by adding a script to fetch the fonts instead of shipping them

This won't work if the machine running gotenberg is not connected to the internet. I am supporting such a configuration.

gulien:
This package has been removed on the `main` branch (or `edge` version of Gotenberg).

The following installed fonts provides metric-compatible equivalents:

```
fonts-crosextra-carlito
fonts-crosextra-caladea 
fonts-liberation 
fonts-liberation2 
```

Issue #1049, opened 2024-11-21
Support arbitrary user id


gulien:
### Discussed in https://github.com/gotenberg/gotenberg/discussions/1042

<div type='discussions-op-text'>

<sup>Originally posted by **leonidas-o** November 12, 2024</sup>
I see a permission denied error in my open shift/okd cluster:
```
{"level":"error","ts":1731426508.9073021,"logger":"api","msg":"convert HTML to PDF: convert to PDF: process first start: start process: run exec allocator: chrome failed to start:\nmkdir: cannot create directory '//.local': Permission denied\ntouch: cannot touch '//.local/share/applications/mimeapps.list': No such file or directory\nchrome_crashpad_handler: --database is required\nTry 'chrome_crashpad_handler --help' for more information.\n","trace":"f031abb3-c554-41ca-ba8c-745b65b73ba2","remote_ip":"x.x.x.x","host":"gotenberg:3000","uri":"/forms/chromium/convert/html","method":"POST","path":"/forms/chromium/convert/html","referer":"","user_agent":"","status":500,"latency":794288676,"latency_human":"794.288676ms","bytes_in":167432,"bytes_out":21}
```

In the past I had already issues with other images, especially when I see something like: https://gotenberg.dev/docs/getting-started/installation#kubernetes
where it says, I should use `runAsUser: 1001`, which you actually don't do in OpenShift.

This could be resolved with the following in the Dockerfile.
"OpenShift: Support arbitrary user ids"
https://docs.openshift.com/container-platform/4.15/openshift_images/create-images.html#use-uid_create-images
```
RUN chgrp -R 0 /home/gotenberg && \
    chmod -R g=u /home/gotenberg
```

Any one here, who could bring up Gotenberg in a restricted cluster like OpenShift/Okd (maybe without creating a custom Gutenberg image)?</div>

leonidas-o:
@gulien  any news on that?

gulien:
No, I did not find the time. Plus, I got a newborn at home, so I don’t have a lot of bandwidth atm 😄

leonidas-o:
oh no, too bad (just the first part), but congratulations for the second part of your message, your newborn. That's amazing!

PR #1143, opened 2025-03-06
address vulnerabilities in docker image

jgoodall:
We are trying to eliminate or reduce vulnerabilities as reported by trivy. This PR replaces the debian docker base image with ubuntu and updates the Dockerfile accordingly.

Tested locally and seemed to work. This removed all high and critical vulnerabilities from trivy:

```
gotenberg/gotenberg:8.17.3-SNAPSHOT (ubuntu 24.04)

Total: 0 (HIGH: 0, CRITICAL: 0)
```


gulien:
Thanks @jgoodall!

Does it increase the size of the image drastically?

jgoodall:
> Does it increase the size of the image drastically?

I replaced the base image with `alpine`, reducing the size by about 1/3. There are no vulnerabilities reported now. I also changed the test image to work properly.

The biggest question is the fonts. I am pretty sure I got all of them; I tried to confirm by looking at the source files and making sure all the true type fonts were installed in the docker image.

gulien:
Hey @jgoodall!

Thanks for your work! 

Having the size reduced by 1/3 is really nice 👍 

A few remarks though:

* I don't see support for the `i386` architecture, at least for Chromium.
* Some dependencies like LibreOffice are less up-to-date than Debian ones (unless using `edge` channel).
* The fonts part feels a bit cumbersome to be honest.

Overall, this is not a minor change, and I don't see myself merging this until Gotenberg 9, which feels a bit early right now 😄

jgoodall:
I am sure we can figure out a way to install chromium on `i386` if that is still needed (maybe from a PPA). I am not sure how to make the fonts less cumbersome, other than maybe hiding that in a script...

I would be happy to revisit as you get closer to releasing v9 - presumably LibreOffice is more up-to-date by then!

ChocoChipset:
Hey @jgoodall ! 
I like the intention of your change, but being such a large change, it introduces a lot of risk. 

Any chance of breaking this into smaller bits? For example: downloading the fonts in their own phase and adding a cache was a really good improvement and is less controversial than changing the whole base image. 

(fiy, I'm not a maintainer)

jgoodall:
I think the answer to your question @ChocoChipset is yes, but i am not really sure what you are suggesting. Right now as much as possible, it just pulls files (like fonts) from the main image so as to not have to recreate things in the alpine variant; I did that to try to minimize risk by reusing what we know works. 

That said, I am happy to make changes if you have specific feedback!

ChocoChipset:
One more question: what was the difference in startup time between alpine and the current image?

gulien:
I also really like the fact that this variant introduces **0 vulnerabilities**—I think it should become the default in version 9.

@jgoodall, could you provide a bit more detail on the following?

- Actual size reduction (in numbers)
- Which tool you used to scan for vulnerabilities, and if possible, a report to illustrate
- Ideally, could you break this down per platform? (e.g., `linux/amd64`, `linux/arm64`, `linux/386`, `linux/arm/v7`)

gulien:
Also, it would be a great opportunity to work on (although it could be done on a separate PR):

- https://github.com/gotenberg/gotenberg/issues/1049
- https://github.com/gotenberg/gotenberg/issues/1101

effyteva:
Hi @gulien ,

Are there any plans to maintain Dockerfiles for additional base image variants, such as Alpine or Ubuntu?

We're currently receiving critical vulnerability alerts related to Debian 13, and we'd prefer to avoid maintaining a custom Dockerfile just to use a different base image.

Keep up the great work.

PR #1164, opened 2025-04-02
fix(webhook): content-disposition header

### LINKED ISSUE #1165: Incorrect Webhook Content-Disposition Header
The Content-Disposition Header for webhooks is incorrectly spelled as `attachement` instead of `attachment`.

We use the webhook to convert files to PDF and upload them directly to our s3 bucket, and started seeing the incorrect header when trying to debug inline vs. attachment views of PDFs in chrome.

It would also be nice to have the Content-Disposition header in the extraHttpHeaders override the default webhook Content-Disposition header defined in the middleware. `attachment` is a good default, but some users may want to set that header for the webhook endpoint to avoid having to send the file to a proxy endpoint to change the Content-Disposition.

stephentgrammer:
Fixes the spelling of `attachment` in the default content-disposition header in the webhook middleware. Also allows a custom content-disposition to be specified via the extraHttpHeaders.

Closes #1165

### DIFF
--- a/pkg/modules/webhook/middleware.go
+++ b/pkg/modules/webhook/middleware.go
@@ -239,9 +239,9 @@ func webhookMiddleware(w *Webhook) api.Middleware {
 						}
 
 						headers := map[string]string{
-							echo.HeaderContentType:        http.DetectContentType(fileHeader),
-							echo.HeaderContentLength:      strconv.FormatInt(fileStat.Size(), 10),
-							traceHeader:                   trace,
+							echo.HeaderContentType:   http.DetectContentType(fileHeader),
+							echo.HeaderContentLength: strconv.FormatInt(fileStat.Size(), 10),
+							traceHeader:              trace,
 						}
 
 						// Allow for custom Content-Disposition header.
--- a/pkg/modules/webhook/middleware.go
+++ b/pkg/modules/webhook/middleware.go
@@ -239,10 +239,15 @@ func webhookMiddleware(w *Webhook) api.Middleware {
 						}
 
 						headers := map[string]string{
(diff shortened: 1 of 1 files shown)


pkg/modules/webhook/middleware.go
```diff
@@ -239,12 +239,17 @@ func webhookMiddleware(w *Webhook) api.Middleware {
...
 							echo.HeaderContentType:        http.DetectContentType(fileHeader),
 							echo.HeaderContentLength:      strconv.FormatInt(fileStat.Size(), 10),
 							traceHeader:                   trace,
 						}
 
+						// if Content-Disposition is not set in extraHttpHeaders, we set it.
+						default_disposition := fmt.Sprintf("attachment; filename=%q", ctx.OutputFilename(outputPath))
+						if _, ok := extraHttpHeaders[echo.HeaderContentDisposition]; !ok {
+							headers[echo.HeaderContentDisposition] = default_disposition
+						}
```
gulien, line 251:
```suggestion
						// Allow for custom Content-Disposition header.
						// See https://github.com/gotenberg/gotenberg/issues/1165.
						_, ok := extraHttpHeaders[echo.HeaderContentDisposition]
						if !ok {
							headers[echo.HeaderContentDisposition] = fmt.Sprintf("attachment; filename=%q", ctx.OutputFilename(outputPath))
						}
```

gulien (changes requested):
Some minor changes but otherwise 👍

stephentgrammer:
@gulien I'm sorry, I'm not sure why the lint is failing. I installed go and the linter on my machine and got it to run, but it didn't find any lint problems. Not sure where to take this next! Thanks.

gulien:
@stephentgrammer looks like a formatting issue! Run `golangci-lint fmt`

gulien:
Thanks @stephentgrammer 😊

PR #1167, opened 2025-04-07
only build debug info when log level is debug

ChocoChipset:
The `BuildDebug` function currently accounts for approximately one-third of the CPU time in the `Run()` function at the start of the Gotenberg service.

By building debug information only when the log level is debug, we can cut cold start times by up to 6-10% for all other cases.
I’ve attached profiling flame graphs that show a reduction in startup time from 3ms to 2ms.

While this 1ms improvement may seem minor, in serverless environments like Cloud Run this could represent 1-3s shaved from the 5-10s startup time required by a typical setup of 1 vCPU/1GB memory. 

With BuildDebug:
Screenshot 2025-04-07 at 2 06 24 PM

Without BuilDebug:
Screenshot 2025-04-07 at 2 06 01 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,7 @@ CHROMIUM_CLEAR_CACHE=false
 CHROMIUM_CLEAR_COOKIES=false
 CHROMIUM_DISABLE_JAVASCRIPT=false
 CHROMIUM_DISABLE_ROUTES=false
+GOTENBERG_BUILD_DEBUG_DATA=true
 LIBREOFFICE_RESTART_AFTER=10
 LIBREOFFICE_MAX_QUEUE_SIZE=0
 LIBREOFFICE_AUTO_START=false
@@ -123,6 +124,7 @@ run: ## Start a Gotenberg container
 	--chromium-clear-cookies=$(CHROMIUM_CLEAR_COOKIES) \
 	--chromium-disable-javascript=$(CHROMIUM_DISABLE_JAVASCRIPT) \
 	--chromium-disable-routes=$(CHROMIUM_DISABLE_ROUTES) \
+	--gotenberg-build-debug-data=$(GOTENBERG_BUILD_DEBUG_DATA) \
 	--libreoffice-restart-after=$(LIBREOFFICE_RESTART_AFTER) \
 	--libreoffice-max-queue-size=$(LIBREOFFICE_MAX_QUEUE_SIZE) \
 	--libreoffice-auto-start=$(LIBREOFFICE_AUTO_START) \
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ build: ## Build the Gotenberg's Docker image
 	-f $(DOCKERFILE) $(DOCKER_BUILD_CONTEXT)
 
 GOTENBERG_GRACEFUL_SHUTDOWN_DURATION=30s
+GOTENBERG_BUILD_DEBUG_DATA=true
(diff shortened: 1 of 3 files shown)


gulien (changes requested):
Good catch! I didn’t expect that 1ms would bubble up to a few seconds in Cloud Run environments 🤯

That said, I’d prefer not to tie this feature to the log level.  

The main reason is that we already have a `--api-enable-debug-route` flag for exposing debug data. If we only build the debug data when the log level is set to `debug`, it could break some environments.
  
Someone might want to expose debug data without enabling debug-level logs.

We also can’t reuse the `--api-enable-debug-route` flag here, since it’s part of a specific module—and that module, while unlikely, may not be loaded in some Gotenberg variants.

I think the best solution is to introduce a new flag: `gotenberg-build-debug-data`, with a default value of `true`.  

You can add it just below the `gotenberg-graceful-shutdown-duration` flag.

ChocoChipset:
let me know what yout think, @gulien!

gulien:
Great 👍 

A few things to add now 😄 

1. Update the [Makefile](https://github.com/gotenberg/gotenberg/blob/7fe0e05050cba8960c2ff5fdd136e8f854020b3a/Makefile#L20) by adding a new arg `GOTENBERG_BUILD_DEBUG_DATA`
2. Add [a test](https://github.com/gotenberg/gotenberg/blob/7fe0e05050cba8960c2ff5fdd136e8f854020b3a/test/integration/features/debug.feature#L8) to validate the output of the `/debug` route when no debug data

ChocoChipset:
Added both points! The new test reveals an empty output that I'm not super happy with, but I'm ok with it from my side for now. 

I guess the ideal scenario would be to have all the debug logic in the debug module, but I imagine the debug module doesn't have access to the other modules.

Makefile
```diff
@@ -52,6 +52,7 @@ CHROMIUM_CLEAR_CACHE=false
 CHROMIUM_CLEAR_COOKIES=false
 CHROMIUM_DISABLE_JAVASCRIPT=false
 CHROMIUM_DISABLE_ROUTES=false
+GOTENBERG_BUILD_DEBUG_DATA=true
```
gulien, line 55: Just need to put it below the `GOTENBERG_GRACEFUL_SHUTDOWN_DURATION` arg 😄

Makefile
```diff
@@ -123,6 +124,7 @@ run: ## Start a Gotenberg container
 	--chromium-clear-cookies=$(CHROMIUM_CLEAR_COOKIES) \
 	--chromium-disable-javascript=$(CHROMIUM_DISABLE_JAVASCRIPT) \
 	--chromium-disable-routes=$(CHROMIUM_DISABLE_ROUTES) \
+	--gotenberg-build-debug-data=$(GOTENBERG_BUILD_DEBUG_DATA) \
```
gulien, line 127: Same!

gulien (changes requested):

Makefile
ChocoChipset: np!

gulien:
@ChocoChipset just add the new test scenario after this one: https://github.com/gotenberg/gotenberg/blob/7fe0e05050cba8960c2ff5fdd136e8f854020b3a/test/integration/features/debug.feature#L8 😄

ChocoChipset:
Like this, @gulien ? ^

PR #1179, opened 2025-04-21
add dependabot for mod and docker files

ChocoChipset:
I've added the mod and docker files to dependabot so they are kept to date automatically. 

Initially there is no effect in the Docker files, but this will allow to pin specific versions of all dependencies and update them automatically, hence having more deterministic builds.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,11 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "gomod" 
+    directory: "/"
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "docker"
+    directory: "/build"
+    schedule:
+      interval: "weekly"


gulien (commented):
Hey @ChocoChipset! Thanks for this PR!

I'm not yet convinced about the `gomod` scanning, as I already update all Go dependencies before a release. I feel like it will mostly spam new PRs.

The Dockerfile with pinned version on the other hand is interesting. Do you have an example so that I understand better your end goal? 😄

ChocoChipset:
1. Indeed, it can become a bit spammy. I think a lower frequency like monthly can still add value in case you go out on holidays, etc. 

2. As it is today, running the same Dockerfile at different points in time can produce a different binary: the package manager is looking for the latest versions available at the time the command is ran. 
By pinning the dependency's version (for example: `exiftool=13.27`) and having dependabot, the Dockerfile can explicitly point to the latest version of the dependency, and follow it automatically. This change would make the Dockerfile more explicit on the version and the builds deterministic.

PR #1181, opened 2025-04-25
Group direct and indirect dependencies

SuperSandro2000:
Generated by combining all dependices into one require group and running go mod tidy.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,42 +4,30 @@ go 1.24.0
 
 require (
 	github.com/alexliesenfeld/health v0.8.0
-	github.com/andybalholm/brotli v1.1.1 // indirect
 	github.com/barasher/go-exiftool v1.10.0
 	github.com/chromedp/cdproto v0.0.0-20250416210000-d7e4d624041a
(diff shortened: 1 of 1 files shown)


gulien:
Hello @SuperSandro2000,

What's the purpose of such organisation?

SuperSandro2000:
That's the standard go sorting. Before it was kinda random.

gulien:
Indeed, but I was using the `go get` command to add dependencies, plus `go mod tidy`.
Where can I find more information about this standard? Is there a tool for sorting them?

SuperSandro2000:
No idea why go get was doing this. Maybe because of the changes done to it in the last years?

Also no idea. This is just the default you get when you have an emtpy go.mod and do go mod tidy.

gulien:
Thanks for this PR @SuperSandro2000! I'll find a way to automatize such behavior, but that obviously not a priority 😄

SuperSandro2000:
I don't think you need to automatise anything here. go will keep that format and I have never needed to sort anything unless before hand it was sorted in some arbitrary groups.

PR #1184, opened 2025-04-26
add closePageAfterConvert option for close context page

RandalTeng:


RandalTeng:
@gulien You see, I just added it as an option. Chromium seems to have an automatic way to reduce memory usage, so I simply added a choice for users to decide whether they want to manually reduce memory.

Of course, setting the default behavior to always auto-close would work, but I think that might feel a bit too aggressive for users.

What do you think it?

RandalTeng:
OK, if you say so, I will do it later, set this behavior as always-do.

RandalTeng:
@gulien hey, 
I have changed the default behavior of closePageAfterConvert, renamed as `keepPageAfterConvert` with a default value of false, which is more readable.

RandalTeng:
I'm hoping you could get a review. Is there anything else that needs to be done?

gulien (changes requested):
Actually, I think it should be as simple as adding:

```go
page.Close()
```

At the end of:

https://github.com/gotenberg/gotenberg/blob/c78ef8771d87a7b2f79e81d8115e9318ac0c8a13/pkg/modules/chromium/browser.go#L282
---
And:
https://github.com/gotenberg/gotenberg/blob/c78ef8771d87a7b2f79e81d8115e9318ac0c8a13/pkg/modules/chromium/browser.go#L306

RandalTeng:
@gulien have done it now.

PR #1186, opened 2025-04-26
change sort of wait delay action and wait expression action, maybe th…

RandalTeng:
…ere is some expression need time to execute.

eg: scroll screen to load all lazy image.


gulien:
Are you using `waitForExpression` for scrolling somehow?

The way I see them, both `waitForExpression` and `waitDelay` are just options to add a custom waiting method for rendering. In other words, you should not use both!

RandalTeng:
hey @gulien.
Actually, in my use case, the `waitForExpression` and `waitDelay` task process as a single thread, I had tried set more like 30 seconds delay, it doesn't work, and than I change the process sort, and it work, so, wo are got here.

And the scrolling page, testing page I will paste it here: https://mp.weixin.qq.com/s/4Sd3EQ9IFTNVUUYo6Y_A3Q
This is a WeChat service account article, I'm not sure it is accessible in outside of China mainland, any article of WeChat is can be test.

PR #1188, opened 2025-04-28
build: pin versions of dependencies

ChocoChipset:
These PR contains two changes: 

1. Adds default arguments for ARG commands and removes them from the .environment file. Dependabot can now keep these up to date. 
2. Pins the versions of dependencies so builds are more deterministic and Dependabot can upgrade them automatically.

### DIFF
--- a/.github/actions/build-test-push/build.sh
+++ b/.github/actions/build-test-push/build.sh
@@ -129,13 +129,7 @@ join() {
 no_arch_tag="$DOCKER_REGISTRY/$DOCKER_REPOSITORY:$version"
 
 cmd="docker buildx build \
-    --build-arg GOLANG_VERSION=$GOLANG_VERSION \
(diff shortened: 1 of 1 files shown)


ChocoChipset:
There were a few dependencies that I couldn't pin, but we have something to start with!

gulien:
Hey @ChocoChipset! 👋 

I'm not convinced this is the way I want to do it actually.

The `.env` file helps centralizing values that are used in different places.

For instance:

* In the [Makefile](https://github.com/gotenberg/gotenberg/blob/9a38f1621917c6e0ba15735a66f4086991e82da2/Makefile#L1)
* In the [GitHub Actions](https://github.com/gotenberg/gotenberg/blob/9a38f1621917c6e0ba15735a66f4086991e82da2/.github/actions/build-test-push/build.sh#L8)

PR #1190, opened 2025-05-03
build: pin versions of dependencies

ChocoChipset:
These PR contains two changes: 

1. Adds default arguments for ARG commands and removes them from the .environment file. Dependabot can now keep these up to date. 
2. Pins the versions of dependencies so builds are more deterministic and Dependabot can upgrade them automatically.


ChocoChipset:
There were a few dependencies that I couldn't pin, but we have something to start with!

Something strange happened with the previous PR. Opening it again. 

I didn't realize .env was also used in the GitHub actions, so I updated it too.

gulien:
Hey @ChocoChipset 👋 

Sorry for the delay, I’ve been quite busy at home the last few days.

I wonder if these changes are working with the multi-arch builds? For instance, a dependency may be available at a given version on `amd64` but not `arm/v7`.

ChocoChipset:
Seems like that's the case. Let me retry with a smaller change.

ChocoChipset:
Actually! the error is because the Docker Hub secret can't be exposed in jobs from PRs created outside of the Gutenberg/gotenberg repository.

gulien:
@ChocoChipset could you do the same work but for the `next` branch? It now make the alpine variant the default one 😄

ChocoChipset:
Happy to help!

ChocoChipset:
I suspect it's the dependencies making the image bigger and ending up the same size. I still think alpine is a better image in this case because of its minimalistic approach. In theory, it has less surface area which would be better for security.

ChocoChipset:
Here is the PR against next:
https://github.com/gotenberg/gotenberg/pull/1207

gulien:
@ChocoChipset I've updated the CI on `main` so that it allows to build and run tests from fork!

build/Dockerfile
```diff
@@ -242,11 +247,11 @@ RUN \
     # See https://github.com/gotenberg/gotenberg/pull/273.
     curl -o /usr/bin/pdftk-all.jar "https://gitlab.com/api/v4/projects/5024297/packages/generic/pdftk-java/$PDFTK_VERSION/pdftk-all.jar" &&\
     chmod a+x /usr/bin/pdftk-all.jar &&\
-    echo '#!/bin/bash\n\nexec java -jar /usr/bin/pdftk-all.jar "$@"' > /usr/bin/pdftk && \
+    printf '#!/bin/bash\n\nexec java -jar /usr/bin/pdftk-all.jar "$@"' > /usr/bin/pdftk && \
```
gulien, line 250: what's the reason for this change? 😄

build/Dockerfile
```diff
@@ -242,11 +247,11 @@ RUN \
...
     chmod +x /usr/bin/pdftk &&\
     apt-get update -qq &&\
     apt-get upgrade -yqq &&\
-    DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends qpdf exiftool &&\
+    DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends qpdf=11.3.0-1+deb12u1 exiftool &&\
```
gulien, line 254: No version found for exiftool?

build/Dockerfile
```diff
@@ -105,18 +106,22 @@ ENV PATH="/opt/java/bin:${PATH}"
...
-ARG PDFTK_VERSION
+ARG GOTENBERG_VERSION=snapshot
+ARG GOTENBERG_USER_GID=1001
+ARG GOTENBERG_USER_UID=1001
+
+# See https://github.com/googlefonts/noto-emoji/releases.
+ARG NOTO_COLOR_EMOJI_VERSION=v2.047
+
+# See https://gitlab.com/pdftk-java/pdftk/-/releases - Binary package.
+ARG PDFTK_VERSION=v3.3.3
```
gulien, line 117:
```suggestion
ARG GOTENBERG_VERSION=snapshot
ARG GOTENBERG_USER_GID=1001
ARG GOTENBERG_USER_UID=1001
# See https://github.com/googlefonts/noto-emoji/releases.
ARG NOTO_COLOR_EMOJI_VERSION=v2.047
# See https://gitlab.com/pdftk-java/pdftk/-/releases - Binary package.
ARG PDFTK_VERSION=v3.3.3
```

build/Dockerfile
```diff
@@ -131,7 +136,7 @@ RUN \
     # Note: tini is a helper for reaping zombie processes.
     apt-get update -qq &&\
     apt-get upgrade -yqq &&\
-    DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends curl gnupg tini python3 &&\
+    DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends curl gnupg=2.2.40-1.1 tini=0.19.0-1 python3=3.11.2-1+b1 &&\
```
gulien, line 139: No version found for curl?

build/Dockerfile
```diff
@@ -198,16 +203,16 @@ RUN \
     /bin/bash -c \
     'set -e &&\
     if [[ "$(dpkg --print-architecture)" == "amd64" ]]; then \
-      curl https://dl.google.com/linux/linux_signing_key.pub | apt-key add - &&\
-      echo "deb http://dl.google.com/linux/chrome/deb/ stable main" | tee /etc/apt/sources.list.d/google-chrome.list &&\
-      apt-get update -qq &&\
-      apt-get upgrade -yqq &&\
-      DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends --allow-unauthenticated google-chrome-stable &&\
-      mv /usr/bin/google-chrome-stable /usr/bin/chromium; \
+    curl https://dl.google.com/linux/linux_signing_key.pub | apt-key add - &&\
```
gulien, line 206: I wonder if we should switch to Chromium for `amd64` 🤔  Nowadays the package is up-to-date with Chrome stable (it wasn't the case before).

gulien (changes requested):
Thanks @ChocoChipset! Added a few comments / suggestions 👍

build/Dockerfile
ChocoChipset: It was flagged by my linter:
"echo won't expand escape sequences. Consider printf."
https://github.com/koalaman/shellcheck/wiki/SC2028

build/Dockerfile
ChocoChipset: I had problems finding the right one for all architectures. Can submit it later on a new PR.

build/Dockerfile
ChocoChipset: Same: didn't find the right one yet for all archs.

ChocoChipset:
Oh! Strange, it says "No conflicts with base branch" here.

gulien:
Thanks @ChocoChipset! Yeah somehow it works now 🤷‍♂️

SuperSandro2000:
Won't this just needlessly brick the docker build as soon as any pinned package gets a new verison uploaded, also for security updates, etc?

gulien:
@SuperSandro2000 AFAIK, we have a GitHub action that should open PRs when deps are updated upstream.

Issue #1144, opened 2025-03-06
Address vulnerabilities in docker image


jgoodall:
We are trying to eliminate or reduce vulnerabilities as reported by trivy. I created a [PR](https://github.com/gotenberg/gotenberg/pull/1143) to eliminate all of the high and critical vulnerabilities reported by trivy by replacing the base image in the Dockerfile. I tested locally and for my use case, everything worked as before.

PR #1206, opened 2025-05-12
Version 9

gulien:
- [ ] Alpine as default - https://github.com/gotenberg/gotenberg/issues/1144
- [x] https://github.com/gotenberg/gotenberg/issues/1049
- [x] https://github.com/gotenberg/gotenberg/issues/1101


gulien:
We have to drop support for `i386` architectures, as there is no Chromium package.

We could still make it work for `arm/v7`, but with JRE 8 instead of 17.

@stumpylog sorry for the direct ping, but any idea if this will impact your users significantly?

stumpylog:
You mean for paperless?  We haven't supported i386 ever and dropped armv7 support a while ago as well

gulien:
> You mean for paperless? We haven't supported i386 ever and dropped armv7 support a while ago as well

Yes for paperless! Thanks for the quick reply!

ChocoChipset:
@gulien, regarding the fonts with the problematic license. Maybe they could be downloaded by a “fonts” module? This circumvents the distribution restriction.

ChocoChipset:
And, what would you think of adding pdfcpu as a library instead of a binary?

Using a binary requires IO, which is super expensive for Cloud Run. :(

gulien:
I was previously using it as a library, but the CLI interface has proven to be more stable.  
It also made process management easier—similar to tools like PDFtk or QPDF. 😄

jgoodall:
@gulien - I pushed changes that merged latest `main` and did updates to alpine image to update golang and alpine and pdfcpu.

PR #1207, opened 2025-05-13
build: pin dependencies' versions on alpine build

ChocoChipset:


ChocoChipset:
The fonts are missing, but I will submit it in another PR.

Issue #645, opened 2023-05-15
PDF merge protected


ytilotti:
I have an issue with a specific pdf (Kbis of https://monidenum.fr/) on PDF Engine merge.
The pdf generated can't be open! I see that the pdf seems protected in edit. It's not possible to merge pdf protected?

gulien:
@ytilotti It makes sense because the PDF engines are not able to open a protected PDF. A workaround would be to send an unprotected version of the PDF you want to merge.

ytilotti:
I don't know if it is possible to check that a downloaded pdf file is protected.

gulien:
What do you mean by snapshot? 😅

ytilotti:
> What do you mean by snapshot? 😅

A print in pdf of the pdf 😊

gulien:
My first guess would be that it won't work either, but you should try and let me know 😊

ytilotti:
For the moment I change my approach with a File Validator (Symfony) on my app to check if the pdf uploaded is protected with /Encrypt.
But why do not use https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-printToPDF @gulien if /Encrypt is present in the file? Maybe a method specific or a param forced.

Issue #453, opened 2022-04-27
Feature request: Add permission passwords to PDF


will2877:
Would it be possible to add permission passwords to the generated PDFs? 

https://helpx.adobe.com/acrobat/using/securing-pdfs-passwords.html

janswist:
I'm also interested in that feature! Any update if it's coming (or how to extract it from existing modules if they are already there) @gulien ?

danielkorytek:
@gulien Any update?

gulien:
@janswist @gulien I'd like to implement it indeed, but I do not have enough free time on my side right now. In other words, I can only do minimal maintainability. 😕

PR #1217, opened 2025-05-22
Adding PDF encryption option

thoven87:
* LibreOffice
* QPDF
* PDFtk
* pdfcpu

### DIFF
--- a/pkg/gotenberg/pdfengine.go
+++ b/pkg/gotenberg/pdfengine.go
@@ -23,6 +23,10 @@ var (
 	// ErrPdfEngineMetadataValueNotSupported is returned when a metadata value
 	// is not supported.
 	ErrPdfEngineMetadataValueNotSupported = errors.New("metadata value not supported")
+
+	// ErrPdfPasswordProtectionNotSupported is returned when password protection
+	// is not supported by the PDF engine.
+	ErrPdfPasswordProtectionNotSupported = errors.New("password protection not supported")
 )
 
 const (
@@ -113,6 +117,12 @@ type PdfEngine interface {
 
 	// WriteMetadata writes the metadata into a given PDF file.
 	WriteMetadata(ctx context.Context, logger *zap.Logger, metadata map[string]interface{}, inputPath string) error
+
+	// ProtectWithPassword adds password protection to a PDF file.
+	// The userPassword is required to open the document.
+	// The ownerPassword provides full access to the document.
+	// If the ownerPassword is empty, it defaults to the userPassword.
+	ProtectWithPassword(ctx context.Context, logger *zap.Logger, inputPath, outputPath string, userPassword, ownerPassword string) error
 }
 
 // PdfEngineProvider offers an interface to instantiate a [PdfEngine].
--- a/pkg/modules/libreoffice/pdfengine/pdfengine.go
+++ b/pkg/modules/libreoffice/pdfengine/pdfengine.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 
 	"go.uber.org/zap"
 
@@ -89,6 +90,61 @@ func (engine *LibreOfficePdfEngine) ReadMetadata(ctx context.Context, logger *za
 // WriteMetadata is not available in this implementation.
 func (engine *LibreOfficePdfEngine) WriteMetadata(ctx context.Context, logger *zap.Logger, metadata map[string]interface{}, inputPath string) error {
 	return fmt.Errorf("write PDF metadata with LibreOffice: %w", gotenberg.ErrPdfEngineMethodNotSupported)
+}
+
+// ProtectWithPassword adds password protection to a PDF file using LibreOffice.
+func (engine *LibreOfficePdfEngine) ProtectWithPassword(ctx context.Context, logger *zap.Logger, inputPath, outputPath string, userPassword, ownerPassword string) error {
+	if userPassword == "" {
+		return errors.New("user password cannot be empty")
+	}
+
+	// LibreOffice can set PDF passwords during conversion
+	opts := api.DefaultOptions()
+	
+	// Configure PDF security options
+	// LibreOffice uses the same options as during export to PDF
+	err := engine.unoApi.Pdf(ctx, logger, inputPath, outputPath, opts)
+	if err != nil {
+		return fmt.Errorf("protect PDF with password using LibreOffice: %w", err)
+	}
+
+	// After LibreOffice generates the PDF, we'll use QPDF to add the password
+	// because LibreOffice doesn't support adding passwords to existing PDFs through command line
+	
+	// Check if QPDF binary exists
+	qpdfBinPath, ok := os.LookupEnv("QPDF_BIN_PATH")
+	if !ok {
+		return errors.New("QPDF_BIN_PATH environment variable is not set")
+	}
+	
+	// If owner password is not provided, use the user password as owner password
+	if ownerPassword == "" {
+		ownerPassword = userPassword
+	}
+	
+	// Use QPDF to apply password protection
+	var args []string
+	args = append(args, outputPath)
+	args = append(args, "--encrypt", userPassword, ownerPassword, "128", "--use-aes=y", "--")
+	args = append(args, outputPath + ".protected")
+	
+	cmd, err := gotenberg.CommandContext(ctx, logger, qpdfBinPath, args...)
+	if err != nil {
+		return fmt.Errorf("create command: %w", err)
+	}
+	
+	_, err = cmd.Exec()
+	if err != nil {
+		return fmt.Errorf("protect PDF with QPDF: %w", err)
+	}
+	
+	// Replace the original file with the protected one
+	err = os.Rename(outputPath + ".protected", outputPath)
+	if err != nil {
+		return fmt.Errorf("replace original file with protected one: %w", err)
+	}
+	
+	return nil
 }
 
 // Interface guards.
--- a/pkg/modules/pdfengines/routes.go
+++ b/pkg/modules/pdfengines/routes.go
@@ -254,6 +254,25 @@ func WriteMetadataStub(ctx *api.Context, engine gotenberg.PdfEngine, metadata ma
 	return nil
 }
 
+// ProtectPdfWithPasswordStub adds password protection to PDF files.
+func ProtectPdfWithPasswordStub(ctx *api.Context, engine gotenberg.PdfEngine, userPassword, ownerPassword string, inputPaths []string) ([]string, error) {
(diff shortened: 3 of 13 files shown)


thoven87:
this PR should close #645 and #453

pkg/modules/pdfengines/routes.go
```diff
@@ -254,6 +254,25 @@ func WriteMetadataStub(ctx *api.Context, engine gotenberg.PdfEngine, metadata ma
...
+		outputPaths[i] = ctx.GeneratePath(".pdf")
+
+		err := engine.ProtectWithPassword(ctx, ctx.Log(), inputPath, outputPaths[i], userPassword, ownerPassword)
+		if err != nil {
+			return nil, fmt.Errorf("protect PDF '%s' with password: %w", inputPath, err)
+		}
+	}
+
+	return outputPaths, nil
+}
```
gulien, line 274: We should keep the filename instead of generating a new one.

Is it possible to use the inputs in the engines instead? For instance, this is what has been done for flattening PDF:

```go
unc (engine *QPdf) Flatten(ctx context.Context, logger *zap.Logger, inputPath string) error {
	var args []string
	args = append(args, inputPath)
	args = append(args, "--generate-appearances")
	args = append(args, "--flatten-annotations=all")
	args = append(args, "--replace-input")
	args = append(args, engine.globalArgs...)
```

test/integration/features/pdfengines_password.feature
```diff
@@ -0,0 +1,103 @@
...
+      | files                     | testdata/page_1.pdf | file   |
+      | files                     | testdata/page_2.pdf | file   |
+      | userPassword              | test123             | field  |
+      | Gotenberg-Output-Filename | protected           | header |
+    Then the response status code should be 200
+    Then the response header "Content-Type" should be "application/zip"
+    Then there should be the following file(s) in the response:
+      | protected.zip |
+    Then the "protected.zip" archive should contain 2 file(s)
+    Then the "protected.zip" archive should contain password protected PDF file(s)
```
gulien, line 85:
```suggestion
    Then there should be 2 PDF(s) in the response
    Then the response PDF(s) should be password protected
```

test/integration/scenario/scenario.go
```diff
@@ -839,6 +839,116 @@ func (s *scenario) thePdfsShouldBeFlatten(ctx context.Context, kind, should stri
...
+	
+	output, err := execCommandInIntegrationToolsContainer(context.Background(), cmd, filePath)
+	
+	// If there's no error, the PDF is not password protected
+	if err == nil && !strings.Contains(output, "password") && !strings.Contains(output, "encrypted") {
+		return fmt.Errorf("expected PDF %s to be password protected, but it is not", filename)
+	}
+	
+	return nil
+}
```
gulien, line 860:
```suggestion
```

test/integration/scenario/scenario.go
```diff
@@ -839,6 +839,116 @@ func (s *scenario) thePdfsShouldBeFlatten(ctx context.Context, kind, should stri
...
+	if err != nil {
+		return fmt.Errorf("parse file count: %w", err)
+	}
+	
+	if fileCount != expectedCount {
+		return fmt.Errorf("expected %d file(s) in archive %s, but found %d", expectedCount, archiveFilename, fileCount)
+	}
+	
+	return nil
+}
```
gulien, line 950:
```suggestion
```

test/integration/scenario/scenario.go
```diff
@@ -874,6 +984,9 @@ func InitializeScenario(ctx *godog.ScenarioContext) {
 	ctx.Then(`^the "([^"]*)" PDF should have (\d+) page\(s\)$`, s.thePdfShouldHavePages)
 	ctx.Then(`^the "([^"]*)" PDF (should|should NOT) be set to landscape orientation$`, s.thePdfShouldBeSetToLandscapeOrientation)
 	ctx.Then(`^the "([^"]*)" PDF (should|should NOT) have the following content at page (\d+):$`, s.thePdfShouldHaveTheFollowingContentAtPage)
+	ctx.Then(`^the "([^"]*)" PDF should be password protected$`, s.thePdfShouldBePasswordProtected)
```
gulien, line 987:
```suggestion
```

test/integration/scenario/scenario.go
```diff
@@ -874,6 +984,9 @@ func InitializeScenario(ctx *godog.ScenarioContext) {
...
+	ctx.Then(`^the "([^"]*)" archive should contain password protected PDF file\(s\)$`, s.theArchiveShouldContainPasswordProtectedPdfFiles)
+	ctx.Then(`^the "([^"]*)" archive should contain (\d+) file\(s\)$`, s.theArchiveShouldContainFiles)
```
gulien, line 989:
```suggestion
```

test/integration/scenario/scenario.go
```diff
@@ -839,6 +839,116 @@ func (s *scenario) thePdfsShouldBeFlatten(ctx context.Context, kind, should stri
...
+	
+	if !foundProtectedPdf {
+		return fmt.Errorf("no password protected PDF files found in archive %s", archiveFilename)
+	}
+	
+	return nil
+}
+
+// theArchiveShouldContainFiles checks if an archive contains the expected number of files.
+func (s *scenario) theArchiveShouldContainFiles(archiveFilename string, expectedCount int) error {
```
gulien, line 923: 😄  Try to replace this with an equivalent of:


```go
func (s *scenario) thePdfsShouldBeFlatten(ctx context.Context, kind, should string) error {
```

pkg/gotenberg/pdfengine.go
```diff
@@ -113,6 +117,12 @@ type PdfEngine interface {
 
 	// WriteMetadata writes the metadata into a given PDF file.
 	WriteMetadata(ctx context.Context, logger *zap.Logger, metadata map[string]interface{}, inputPath string) error
+
+	// ProtectWithPassword adds password protection to a PDF file.
+	// The userPassword is required to open the document.
+	// The ownerPassword provides full access to the document.
+	// If the ownerPassword is empty, it defaults to the userPassword.
+	ProtectWithPassword(ctx context.Context, logger *zap.Logger, inputPath, outputPath string, userPassword, ownerPassword string) error
```
gulien, line 125:
```suggestion
	AddPassword(ctx context.Context, logger *zap.Logger, inputPath, outputPath string, userPassword, ownerPassword string) error
```

To reflect the route naming?

test/integration/features/pdfengines_password.feature
```diff
@@ -0,0 +1,103 @@
...
+
+  Scenario: POST /forms/pdfengines/add-password with password engines that don't support password protection
+    Given I have a Gotenberg container with the following environment variable(s):
+      | PDFENGINES_PASSWORD_ENGINES | exiftool |
+    When I make a "POST" request to Gotenberg at the "/forms/pdfengines/add-password" endpoint with the following form data and header(s):
+      | files                     | testdata/page_1.pdf | file   |
+      | userPassword              | test123             | field  |
+      | Gotenberg-Output-Filename | protected           | header |
+    Then the response status code should be 500
+    Then the response body should contain "password protection not supported"
```
gulien, line 103: Misses a few test that should just be c/c with minor changes 😄 Check `pdfengines_flatten.feature`

gulien (changes requested):
Great PR @thoven87! Thank you very much!

I added a few request for changes and suggestions in my review, but nothing fancy 😉 

Also, it could be great to add this feature for the following routes:

* `/forms/chromium/convert`
* `/forms/pdfengines/merge`
* `/forms/pdfengines/split`

And update their corresponding tests.

You can do that in another PR if you prefer ofc.

Later on, would you also like to work on password "unprotection" (to close #645) ? 😄

pkg/modules/libreoffice/pdfengine/pdfengine.go
```diff
@@ -91,6 +92,61 @@ func (engine *LibreOfficePdfEngine) WriteMetadata(ctx context.Context, logger *z
 	return fmt.Errorf("write PDF metadata with LibreOffice: %w", gotenberg.ErrPdfEngineMethodNotSupported)
 }
 
+// ProtectWithPassword adds password protection to a PDF file using LibreOffice.
+func (engine *LibreOfficePdfEngine) ProtectWithPassword(ctx context.Context, logger *zap.Logger, inputPath, outputPath string, userPassword, ownerPassword string) error {
```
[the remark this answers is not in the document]
thoven87: > We should avoid using LibreOffice when possible for PDF to PDF transformations, as it's a bit hacky 😄 just return a `gotenberg.ErrPdfPasswordProtectionNotSupported` here!

Sounds good! I will drop the LibreOffice implementation.

pkg/gotenberg/pdfengine.go
thoven87: I am wondering if the route name should be encrypt instead of `password/add` and later on `password/remove`. What do you think @gulien?

gulien: Good idea! Let's rename it to `Encrypt`!

thoven87:
> Almost there 😄 !
> 
> FYI, the encrypt feature should only be available for the following routes:
> 
> * `/forms/chromiums/convert/*`
> * `/forms/libreoffice/convert`
> * `/forms/pdfengines/merge`
> * `/forms/pdfengines/split`
> * `/forms/pdfengines/encrypt`
> 
> I've seen this feature enabled for other routes in your integration tests 😉
> 
> Also, regarding these integrations tests:
> 
> ```
>     Then there should be the following file(s) in the response:
>       | encrypted.zip |
>     Then the "encrypted.zip" archive should contain 2 file(s)
>     Then the "encrypted.zip" archive should contain encrypted PDF file(s)
> ```
> 
> Should be replaced with:
> 
> ```
>     Then the response header "Content-Type" should be "application/zip"
>     Then there should be 2 PDF(s) in the response
>     Then the response PDF(s) should be encrypted
> ```
> 
> The first two scenarios already exists, but you just have to create the last one 😄
> 
> For the `POST /forms/pdfengines/encrypt` feature, you should also add the remaining tests like `Route Disabled`, `Download From`, `Webhook`, etc.

Thanks for the feedback @gulien I have not written much Go before. I appreciate the hints. Would a user not want to flatten and encrypt at the same time? I need some hint to the remaining routes / webhook tests.

gulien:
Your PR looks pretty good IMO, you've followed the Gotenberg's patterns greatly 👍 

I think it makes sense for merging / splitting PDFs, but less for others. Also, if all PDF engines routes accept the same form fields, at then end the routes are all the same 😄 

> I need some hint to the remaining routes / webhook tests.

Just clean your tests with what I mentioned, I'll add the remaining ones 😉

Issue #1230, opened 2025-06-07
Fonts not loading in 8.21.0


scott-the-programmer:
Hey! Congrats on the newborn! 🍼 

I noticed that fonts aren't being embedded correctly with the latest version. I believe this is actually due to a bad commit in chromium as I was able to rebuild the image again locally and get the correct fonts.

Both the PDFs below are generated from the example pdf https://sparksuite.github.io/simple-html-invoice-template/ on forms/chromium/convert/url

8.20.1:

![Image](https://github.com/user-attachments/assets/be2d57f6-353e-4795-9cfb-0a82bfd405c3)

8.21.0:

![Image](https://github.com/user-attachments/assets/b71bcd97-4f2b-48cf-ad56-aa9536d3e03c)

I suspect this also relates to issue #1218,

I think all that is required here is to do a rebuild and release 😁.

# Workaround

For those affected, simply pinning to 8.20.1 works fine i.e. `gotenberg/gotenberg:8.20.1`

scott-the-programmer:
I should add that the Chrome version that worked locally was `137.0.7151.68`

gulien:
Hey @scott-the-programmer 👋 

Thanks! 👶

Could you test `gotenberg/gotenberg:edge` and tells me if it works? 😁

scott-the-programmer:
@gulien Looks good! 

![Image](https://github.com/user-attachments/assets/e053b910-0a4b-4f88-a804-e30650dd77b5)

gulien:
Version `8.21.1` should fix the issue 😄 thanks @scott-the-programmer!

muratbinerbay:
Hi @gulien,

We had version `8.21.1` pinned due to the fonts issue and everything worked.

Tried updating to `8.24` today and custom fonts don't load anymore.
This is how we add fonts:
```
COPY .docker/gotenberg/bootstrap-icons.woff /usr/local/share/fonts/
COPY .docker/gotenberg/bootstrap-icons.woff2 /usr/local/share/fonts/
```

Also tried `8.23` and `8.22` and the fonts don't load there either.

Also added the code below to our Dockerfile, but didn't help. We haven't needed that until now.

```
RUN fc-cache -fv \
 && chgrp -R 0 /usr/local/share/fonts \
 && chmod -R g=u /usr/local/share/fonts
```

Is anyone else having a similar issue with fonts again?

scott-the-programmer:
[...]

PR #1231, opened 2025-06-07
add docker build arg to pin chromium/chrome version

scott-the-programmer:
Closes https://github.com/gotenberg/gotenberg/issues/1230 

This isn't really necessary since building against the latest chromium will fix issue #1230, however I thought it would be a neat bonus to be able to control the chromium version as a build arg. This is so that someone can control the version of chromium against a version of gotenberg

Feel free to disregard!


scott-the-programmer:
Nevermind! Looks like the repository doesn't keep past versions

PR #1233, opened 2025-06-08
Add support for having sync upload with webhook url

### LINKED ISSUE #1224: Feature Request: Synchronous file upload option for webhook upload url
Gotenberg's webhook feature does async file uploads to a given URL and sends back a 204 No Content response. But in something like AWS Lambda, where the function shuts down right after responding, the async setup doesn’t work. Adding a sync option that handles the upload in the same request while still giving a 204 response would be super helpful for these cases.

I know this might tweak the usual webhook vibe in Gotenberg (or webhooks in general), but it’s a solid use case. 

What do you think about adding this? If it’s a go, I’m happy to contribute.

ynarwal:
Closes #1224

### DIFF
--- a/pkg/modules/webhook/middleware.go
+++ b/pkg/modules/webhook/middleware.go
@@ -20,11 +20,74 @@ import (
 	"github.com/gotenberg/gotenberg/v8/pkg/modules/api"
 )
 
+type SendOutputFileParams struct {
+	Ctx              *api.Context
+	OutputPath       string
+	ExtraHttpHeaders map[string]string
+	TraceHeader      string
+	Trace            string
+	Client           *client
+	HandleError      func(error)
+}
+
 func webhookMiddleware(w *Webhook) api.Middleware {
 	return api.Middleware{
 		Stack: api.MultipartStack,
 		Handler: func() echo.MiddlewareFunc {
 			return func(next echo.HandlerFunc) echo.HandlerFunc {
+				sendOutputFile := func(params SendOutputFileParams) {
+					outputFile, err := os.Open(params.OutputPath)
+					if err != nil {
+						params.Ctx.Log().Error(fmt.Sprintf("open output file: %s", err))
+						params.HandleError(err)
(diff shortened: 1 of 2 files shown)


pkg/modules/webhook/middleware.go
```diff
@@ -20,11 +20,74 @@ import (
 	"github.com/gotenberg/gotenberg/v8/pkg/modules/api"
 )
 
+type SendOutputFileParams struct {
```
gulien, line 23:
```suggestion
type sendOutputFileParams struct {
```

Keep it private 😄

pkg/modules/webhook/middleware.go
```diff
@@ -173,10 +235,49 @@ func webhookMiddleware(w *Webhook) api.Middleware {
 						}
 					}
 
-					w.asyncCount.Add(1)
+					webhookSync := strings.ToLower(c.Request().Header.Get("Gotenberg-Webhook-Sync")) == "true"
```
gulien, line 238:
```suggestion
					webhookSync, err := strconv.ParseBool(c.Request().Header.Get("Gotenberg-Webhook-Sync"))
```

`ParseBool` is able to parse more "boolean" strings.

gulien (changes requested):
Looks great!

Don't forget to add a test scenario in https://github.com/gotenberg/gotenberg/blob/ffeb4703fd85666fa5cbd970eb97ce1721d95dfe/test/integration/features/webhook.feature.

pkg/modules/webhook/middleware.go
ynarwal: Yep. I knew that is going to be asked. Done.

ynarwal:
@gulien hey congrats on new born! just following up on upcoming release, do you have any indicative timeline for this be released :)?

gulien:
@ynarwal `gotenberg/gotenberg:edge` should be available with this feature :)

Issue #1214, opened 2025-05-21
Support for Sentry.io exception reporting


peterfarrell:
@gulien would you be open to a PR that adds **optional** support to capture exceptions and send thems to Sentry for error reporting?  For example, GitLab runners for their pipelines offer optional integration with Sentry to error runner exceptions (including stack traces). I did not want to work on this without knowing how receptive a PR on this would be.

lorello:
This sounds really interesting also for us. We already use Sentry for error tracing in our microservices, having it also in gotenberg should be a great plus! Thanks @gulien !

gulien:
Hello @peterfarrell 👋 

I think a better option would be to implement OpenTelemetry instead? 🤔  See #814.

peterfarrell:
G'day @gulien 👋 

Thanks for your response and for mentioning OpenTelemetry in #814.

While OpenTelemetry is valuable for broad observability, our immediate and main focus is on robust exception and error reporting, management, and workflow.

Sentry excels here with:

- **Distributed Tracing:** Sentry's distributed tracing allows us to link exceptions in Gotenberg directly to the originating requests and processes in our applications, providing full end-to-end context.
- **Intelligent Error Grouping:** It automatically groups similar errors into actionable "issues," cutting down noise.
- **Rich Contextual Data:** It provides vital context like detailed stack traces, breadcrumbs, and user info for faster debugging.
- **Streamlined Workflow:** Sentry offers a dedicated UI and tools for triaging, assigning, and resolving issues efficiently.

I realize that OpenTelemetry can capture exception events, but it doesn't offer this specialized level of issue management and contextual analysis. For proactive incident management and developer efficiency, Sentry's focused approach to exceptions is a better fit for our current needs.

[Sentry's Go SDKs can integrate with OpenTelemetry](https://docs.sentry.io/platforms/go/tracing/instrumentation/opentelemetry/) for future performance tracing, but for exception reporting, some type of Sentry integration is our priority. What are your thoughts on this more focused approach?

gulien:
Sure, let's open a PR!

peterfarrell:
@gulien Awesome. I will attempt something in the coming days. I'm not terribly familar with Go but Gemini Code Assist can help with that. 😺 However, code assist can't substitute for good "taste" in Go. I will welcome constructive feedback and revisions to a PR if you are willing to provide it. Looking forward to the challenge!

PR #1237, opened 2025-06-09
feat: initial pass at sentry module, still WIP

peterfarrell:
Sentry is integrated as a Gotenberg module and is disabled unless a Sentry DSN is provided. See #1214 for discussion.

WIP: catching and reporting issues to Sentry

Items still TODO / WIP:

- [ ] Tests (please advise what you find important here)
- [ ] Docs


peterfarrell:
@gulien This draft PR has a module for Sentry modeled after how the Prometheus module was built.  The Sentry module reads the command-line args / env variables and init's the Sentry SDK for use.

Still WIP is reporting errors to Sentry. I'm not very familar with Go (nor Echo) so I thought it would be best to ask your opinion.

It appears that Echo is the framework used for the API module.  Sentry has support for Echo using the Echo middleware.  Does implementing Sentry support using a middleware seem OK to you or do you have another preferred method you'd like me to use?

https://docs.sentry.io/platforms/go/guides/echo/

Assuming that using the Echo middleware is the manner to do this, can you confirm the Gotenberg pattern would be for the Sentry module to create a `middleware.go` with `sentryMiddleware` (akin to the webhook middleware) that enables the Sentry middleware for Echo if Sentry is enabled by the configuration?

gulien (changes requested):
Thanks @peterfarrell! 

Take a look at the `prometheus` module for a complete example:
https://github.com/gotenberg/gotenberg/blob/ffeb4703fd85666fa5cbd970eb97ce1721d95dfe/pkg/modules/prometheus/prometheus.go

It's almost the same module pattern IMO 😄

peterfarrell:
@gulien

1) Resolved suggested changes from you
2) Add manually added Sentry support to capture all 400, 500, and 503 http exceptions that do not bubble up as a panic (panics are auto handled by the Sentry Echo middleware).

Makefile
```diff
@@ -66,6 +66,9 @@ PROMETHEUS_NAMESPACE=gotenberg
 PROMETHEUS_COLLECT_INTERVAL=1s
 PROMETHEUS_DISABLE_ROUTE_LOGGING=false
 PROMETHEUS_DISABLE_COLLECT=false
+SENTRY_DSN=
+SENTRY_SEND_DEFAULT_PII=true
+SENTRY_ENVIRONMENT=production
```
gulien, line 71:
```suggestion
SENTRY_ENVIRONMENT=
```

Should be the default values here 😄 

Also it misses `sentry-flush-timeout`

pkg/modules/api/middlewares.go
```diff
@@ -76,6 +77,17 @@ func httpErrorHandler() echo.HTTPErrorHandler {
...
 		status, message := ParseError(err)
 
+		// Capture non-panic errors and send them to Sentry.
+		// Panics are already handled by the sentryecho middleware.
+		// Send errors to Sentry if the status code indicates a client error (400),
+		// an internal server error (500), or service unavailability (503).
+		// hub.CaptureException(err) sends the event with sentry.LevelError by default.
+		if status == http.StatusBadRequest || status == http.StatusInternalServerError || status == http.StatusServiceUnavailable {
+			if hub := sentryecho.GetHubFromContext(c); hub != nil {
+				hub.CaptureException(err)
```
gulien, line 87: This behavior should actually be implemented in the module directly.

Check for instance:

https://github.com/gotenberg/gotenberg/blob/main/pkg/modules/webhook/middleware.go

pkg/modules/sentry/sentry.go
```diff
@@ -0,0 +1,162 @@
...
+		err := sentry.Init(s.sentryClientOptions)
+		if err != nil {
+			// This error is reported if a dsn was provided but Sentry failed to initialize.
+			return fmt.Errorf("Sentry configuration error: %w", err)
+		}
+
+		s.sentryInitialized = true
+	}
+
+	return nil
```
gulien, line 120:
```suggestion
        if s.sentryClientOptions.Dsn == "" {
            return nil
        }
        
	err := sentry.Init(s.sentryClientOptions)
	if err != nil {
	        // This error is reported if a dsn was provided but Sentry failed to initialize.
	        return fmt.Errorf("Sentry configuration error: %w", err)
        }
        
        s.sentryInitialized = true
		
	return nil
```

pkg/modules/sentry/sentry.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return "integration is disabled - no DSN provided"
+}
+
+// Stop flushes any remaining events.
+func (s *Sentry) Stop(ctx context.Context) error {
+	// Flush buffered events before the program terminates, but only when Sentry client is initialized.
+	// Set the timeout to the maximum duration the program can afford to wait.
+	if s.sentryInitialized {
+		defer sentry.Flush(s.flushTimeout)
```
gulien, line 136:
```suggestion
		defer sentry.Flush(s.flushTimeout)
```

I think we should use the context deadline 🤔

gulien (changes requested):

pkg/modules/sentry/sentry.go
peterfarrell: @gulien thanks for the comment - I don't know what that is or how to do it. Could you please advise or provide a bit more context?

peterfarrell: @gulien are you thinking of something like this where we can compute the remaining time on the deadline? If this was the case, I'd set the default of `flushTimeout` to `10s` and then a minimal positive timeout of `1s` if the context is done or the deadline passed.

Your feedback on this is appreciated.

```go
// Stop flushes any remaining events.
func (s *Sentry) Stop(ctx context.Context) error {
	// Flush buffered events before the program terminates, but only when Sentry client is initialized.
	if s.sentryInitialized {
		// Defer the flush operation. The timeout will be determined based on the context
		// when the deferred function executes.
		defer func(stopCtx context.Context) {
			// Start with the configured Sentry flush timeout as the default/upper bound.
			timeoutForFlush := s.flushTimeout

			// If the Stop context has a deadline, the flush must not exceed
			// the time remaining for that deadline.
			if deadline, ok := stopCtx.Deadline(); ok {
				remainingTimeForContext := time.Until(deadline)
				if remainingTimeForContext < timeoutForFlush {
					// Context offers less time than s.flushTimeout, so use the context's remaining time.
					timeoutForFlush = remainingTimeForContext
				}
			}

			// If the context is already marked as "done" (e.g., canceled, or deadline already passed),
			// or if the calculated timeout ended up being non-positive (deadline passed during Stop execution),
			// then use a very small positive timeout for a best-effort, non-blocking flush.
			if err := stopCtx.Err(); err != nil {
				s.logger.Debug("Sentry Stop: context is already done, using minimal flush timeout.", zap.Error(err))
				timeoutForFlush = 1 * time.Second // Minimal positive timeout
			} else if timeoutForFlush <= 0 {
				s.logger.Debug("Sentry Stop: calculated flush timeout is non-positive (deadline likely passed), using minimal flush timeout.")
				timeoutForFlush = 1 * time.Second // Minimal positive timeout
			}

			s.logger.Debug("Sentry: flushing events", zap.Duration("timeout", timeoutForFlush))
			sentry.Flush(timeoutForFlush)
		}(ctx) // Pass the context to the deferred function's closure
	}

	return nil
}
```

pkg/modules/api/middlewares.go
peterfarrell: @gulien Does something like this work for the `sentry/middlewares.go`? 

Additional items to point out:

- Changed the priority in `sentryPanicMiddleware` to the `HighPriority` to it's available early
- The priority in the `sentryErrorCaptureMiddleware` is set to `LowPriority`

```go
package sentry

import (
	"net/http"

	sentryecho "github.com/getsentry/sentry-go/echo"
	"github.com/labstack/echo/v4"

	"github.com/gotenberg/gotenberg/v8/pkg/modules/api"
)

// sentryPanicMiddleware is the middleware from sentry-go/echo for panic reporting and hub setup.
// It should run with a higher priority to ensure the Sentry Hub is available for other middlewares.
func sentryPanicMiddleware() api.Middleware {
	return api.Middleware{
		Stack:    api.DefaultStack,
		Priority: api.HighPriority, // Ensures Sentry Hub is set up early.
		Handler: sentryecho.New(sentryecho.Options{
			Repanic: true,
		}),
	}
}

// sentryErrorCaptureMiddleware captures specific non-panic errors and sends them to Sentry.
// It should run with a lower priority to catch errors from main handlers.
func sentryErrorCaptureMiddleware() api.Middleware {
	return api.Middleware{
		Stack:    api.DefaultStack,
		Priority: api.LowPriority, // Runs after most handlers but before final error response generation.
		Handler: func(next echo.HandlerFunc) echo.HandlerFunc {
			return func(c echo.Context) error {
				err := next(c)
				if err != nil {
					// Use api.ParseError to determine the HTTP status code this error would generate.
					status, _ := api.ParseError(err)

					if status == http.StatusBadRequest || status == http.StatusInternalServerError || status == http.StatusServiceUnavailable {
						if hub := sentryecho.GetHubFromContext(c); hub != nil {
							hub.CaptureException(err)
						}
					}
				}
				return err // Return the original error to be handled by Echo's main error handler.
			}
		},
	}
}
```

pkg/modules/sentry/sentry.go
peterfarrell: @gulien code added for this suggestion:

https://github.com/gotenberg/gotenberg/pull/1237/files#diff-48f5b89967188c919bf21913391a6283dde27a59b9cf424dfaec728ca7cfd5f3R132

pkg/modules/api/middlewares.go
peterfarrell: @gulien code added  for this:

https://github.com/gotenberg/gotenberg/pull/1237/files#diff-e047d4a64ae437dee633eb25e47e4cb6a01d12c003dfef3a546af14873f0fad9

peterfarrell:
@gulien I've done a fair amount of research and LLM-ing however I'm not able to successful run `make test-integration`.  This is likely because I'm not familiar with how action is performed in Go and I'm willing to ask what may be newbie question.

I do this:

1. `make build`
2. `make test-integration`

This failures with this error:

```shellsession
    create Gotenberg container: create Gotenberg container network: reaper: new reaper: run container: 
    Error response from daemon: {"message":"unable to retrieve auth token: invalid username/password: 
    unauthorized: incorrect username or password"}: failed to create network
```

I'm running Docker Desktop on MacOS.  BTW, `make test-unit` runs successfully.

Would you open to a separate PR that defines a `CONTRIBUTION.md` file and includes pre-reqs and step to be able to do Gotenberg development on your local machine?  I've felt my way through this by looking at the makefile, and trial/error.  It may help folks to do PRs and make the process a little less intimidating. LMK.

gulien:
Hello @peterfarrell,

Sorry for the delay. I’m starting a new position at a new company, plus my newborn, I have 0 free time 🥲

I agree I should add a `CONTRIBUTING.md` 👍

peterfarrell:
> I agree I should add a `CONTRIBUTING.md` 👍

After I learn some of the testing item for this PR, I'm happy to do another PR after this one to start a `CONTRIBUTING.md`.

Lachee:
Would it be possible to add [Trace Propagation/Distribution](https://docs.sentry.io/platforms/go/guides/echo/tracing/trace-propagation/) in this too using the `sentry-trace` and `baggage` headers. 

It should both listen to them on API requests and also emit them when requesting content. That way it will be possible to follow a trace from request to render.

peterfarrell:
@Lachee The current WIP PR uses the Echo middleware for both panic and non-panic issue reporting to Sentry:

https://github.com/gotenberg/gotenberg/pull/1237/files#diff-e047d4a64ae437dee633eb25e47e4cb6a01d12c003dfef3a546af14873f0fad9

The documentation indicates that the Echo middleware uses the `sentry-trace` and `baggage` headers to enable Trace Progagation support:

https://docs.sentry.io/platforms/go/guides/echo/tracing/trace-propagation/

So it appears Trace Progation is already supported due to the use of the Echo middleware. I'm a Go novice so anyone with more knowledge, please chime in.

We'd need to ensure that Gotenberg doesn't strip those headers away and include some basic documentation that your proxy (like HAProxy, etc.) may strip those headers away prior to reaching Gotenberg:

> If you run any JavaScript applications in your distributed system, make sure that those two headers are added to your CORS allowlist and won't be blocked or stripped by your proxy servers, gateways, or firewalls.

peterfarrell:
@gulien wanted to check in to see when you might have bandwidth to review this PR. (understandably you are busy with a new little one)

gulien:
@peterfarrell really sorry! Also starting a new position, so I have no free time to spend on Gotenberg atm!
But I have all PRs in mind, don't worry!

peterfarrell:
@gulien I have sync'ed this branch with main and resolved the conflicts.  Ready for review.  I'd love to resurrect this feature request.

gulien:
Hello @peterfarrell 👋 

Thanks again for this PR! I fisrt want to merge #1436 and then see how to integrate Sentry in the flow. I think a simple propagator would be enough (and maybe making your PR useless 😅😅).

peterfarrell:
@gulien Thanks for the heads-up — I'll hold off on further updates here until #1436 lands.

If the new propagator natively supports trace propagation and distribution (specifically with Sentry), it will likely make this PR overcome by events (OBE).

However, as @Lachee noted earlier, the core requirement for us is the ability to link Gotenberg exception/error events directly to the API requests initiated by our applications. If the propagator doesn't address that specific trace connectivity, we may still need a way to pass that context manually. I'll re-evaluate once the new implementation is available!

gulien:
@peterfarrell the OpenTelemetry integration is now live! Let me know if it meets your requirements!

peterfarrell:
@gulien I'll take a peek -- it looks promising based on the Sentry OTEL Collector documentation:

https://sentry-build-otlp.sentry.dev/using-otel-collector/

peterfarrell:
@gulien I've successfully configured traces and logs from Gotenberg to Sentry via an Otel Collector, but this approach introduces significant operational complexity and functional limitations compared to the native Sentry SDK integration.

### Critical Limitations Identified

**1. Distributed Tracing Context Propagation**
The Otel Collector path cannot propagate incoming distributed tracing headers (`sentry-trace`, `baggage`, `traceparent`) from API requests through to Sentry. This breaks trace continuity across service boundaries and would require modifications to Gotenberg's core request handling to support properly.

**2. Environment Configuration**
Setting the Sentry environment (`production`, `staging`, `dev-abc123`) requires adding a resource processor to the Otel Collector configuration. The SDK handles this natively through simple configuration parameters.

**3. Error Context Fidelity**
When Gotenberg encounters errors, the Otel pipeline provides significantly reduced context compared to the Sentry SDK. The SDK captures:
- Full stack traces with local variables
- Breadcrumbs (sequence of events leading to the error)
- Request context and user information
- Custom tags and fingerprinting

The Otel approach loses much of this enriched error context during translation.

**4. Outbound Request Instrumentation**
Outbound HTTP requests appear in Otel traces but lack distributed tracing headers. The Sentry SDK's automatic instrumentation injects `sentry-trace` and `baggage` headers on outbound requests, enabling full trace propagation across your service mesh.

### Recommendation

The native Sentry SDK integration via Echo middleware provides:
- **Zero-configuration distributed tracing** with automatic header propagation
- **Rich error context** without data loss
- **Simplified deployment** without additional infrastructure (no sidecar containers)
- **Full feature parity** with Sentry's error tracking and performance monitoring

While the Otel Collector excels at **log aggregation and vendor-neutral telemetry routing**, it introduces an impedance mismatch for Sentry's error tracking and distributed tracing features. The SDK integration delivers complete observability with minimal operational overhead.

For Gotenberg's use case, the Sentry SDK provides the most robust and maintainable solution.

peterfarrell:
For those that stumble upon this PR, a reference of how I setup the Otel Collector with Sentry for Gotenberg is on this snippet:

https://gitlab.com/-/snippets/5975695

Hopefully, this will save some folks some time if they wish to use Otel with Sentry.

nikolasdas:
I'm new to Otel, so sorry if this is a bit of a stupid question :D
Does it work if I simply point the `OTEL_*` env vars to the sentry OTLP endpoints? Or do I need the collector running?

peterfarrell:
@nikolasdas You need to setup an Otel collector in order to export it to Sentry.  See this snippet that hows an example setup of Gotenberg with an Otel collector that exports to Sentry:

https://gitlab.com/-/snippets/5975695

nikolasdas:
@peterfarrell thank you for the snippet. Since my paas provider does not support docker compose setups in a single service and I'm trying to avoid adding another one just for the collector, I tried my setup as described above (simply pointing the env vars to sentry).
I do see logs and traces. So it seems that at least some parts are working without a collector

gulien:
I think OpenTelemetry provides everything required for Sentry observability. Thanks all for the investigations!

peterfarrell:
@gulien Please see this comment on the operational complexities and deficiencies of using the Otel only approach:

https://github.com/gotenberg/gotenberg/pull/1237#issuecomment-4183833558

PR #1277, opened 2025-07-19
fix bin extension

PhilMeshkovo:


PhilMeshkovo:
sometimes .bin extension is also used for conversion. 
Libreoffice gives this possibility

gulien:
Hello @PhilMeshkovo! Does it work with this change? Could you add a testing case?

PhilMeshkovo:
Hello @gulien . I added a test and ran it locally, everything is ok
image

gulien:
Hey @PhilMeshkovo,
Thanks a lot for the idea! After giving it some thought, I’m not sure it’s safe enough to include, so I don’t think it’s the right fit for Gotenberg at the moment.

PR #1289, opened 2025-08-05
fix(chromium): Enable GPU support

EyeCantCU:
Chromium 139 is set to deprecate the Swiftshader backend entirely:

https://chromestatus.com/feature/5166674414927872

And 138 doesn't provide fallback without explicitly enabling it, breaking functionality such as WebGL in Gotenberg

Re-enable GPU support to work around this

### DIFF
--- a/pkg/modules/chromium/browser.go
+++ b/pkg/modules/chromium/browser.go
@@ -84,10 +84,6 @@ func (b *chromiumBrowser) Start(logger *zap.Logger) error {
 		chromedp.ExecPath(b.arguments.binPath),
 		chromedp.NoSandbox,
 		// See:
-		// https://github.com/gotenberg/gotenberg/issues/327
-		// https://github.com/chromedp/chromedp/issues/904
-		chromedp.DisableGPU,
-		// See:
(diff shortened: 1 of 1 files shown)


EyeCantCU:
An alternative would be to set `enable-unsafe-swiftshader` moving forward. Though I do believe this is the way to go. The issues raised (that led to GPU support being disabled originally) don't appear to still be relevant

gulien:
Hey @EyeCantCU 👋 

Thanks for this MR!

AFAIK, I disabled GPU because it causes issues on environments with.. no GPU 😄 (k8s clusters etc.).

EyeCantCU:
Hi @gulien! Really appreciate the response. I tested this change in a VM on GCE without GPU access and Chromium didn't fail to start or present any odd behavior that was present in the linked issues. If you'd prefer to retain existing behavior though, we can just enable `--enable-unsafe-swiftshader`

Chromium 139 will break current behavior and upstream builds of Chromium 138 already have without `--enable-unsafe-swiftshader` set

gulien:
Thanks @EyeCantCU!

PR #1352, opened 2025-10-08
fix(deps): revert indirect nwaples/rardecode/v2 bump

omni-htg:
During https://github.com/gotenberg/gotenberg/commit/e5ccb82db84637eb75237cc6b5e951edb8f228e7 there was an update to the `nwaples/rardecode/v2` package from v2.1.1 to v2.2.0 .
Unfortunately, this package is a depended on by `mholt/archives`, and v0.1.4 [does not support bumping rardecode to v2.2.0](https://github.com/mholt/archives/issues/55).
You can see it logged as well on the GitHub Actions runs failing to build.

This PR reverts that change.


gulien:
Great! Thank you very much @omni-htg!

PR #1375, opened 2025-10-28
no need for proxy

Vagif12:


gulien:
Hello @Vagif12,

Could you provide some context? 😄

PR #1376, opened 2025-10-29
Poc shrink image size

gunnar-rydberg:
Removes chromium and pdftk (including Java) to save some space.


gunnar-rydberg:
🤦🤦🤦 Sorry, mixed up my forks...

PR #1446, opened 2026-01-15
Add `waitForSelector` option to Chromium conversions

### LINKED ISSUE #960: Feature Request: wait for selector
Hi, would it be possible to add a "waitForSelector" method ?

Using Playwright, you would do something like 
`const res = await page.waitForSelector(selector)`

Something like this : https://playwright.dev/docs/api/class-elementhandle#element-handle-wait-for-selector

Thanks ! :)

danxmoran:
Closes #960

As an alternative to waiting on an expression, this allows users to wait for a specific node matching a selector to become visible in the HTML / at the remote URL before converting to PDF.

### DIFF
--- a/pkg/modules/chromium/tasks.go
+++ b/pkg/modules/chromium/tasks.go
@@ -521,7 +521,8 @@ func waitForSelectorVisibleBeforePrintActionFunc(logger *zap.Logger, selector st
 		}
 
 		logger.Debug(fmt.Sprintf("wait until '%s' is visible before print", selector))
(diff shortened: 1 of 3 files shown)


danxmoran:
@gulien thanks for the quick review - comments addressed!

gulien:
Would you mind updating the documentation with this new feature? 😄

https://github.com/gotenberg/gotenberg.dev

danxmoran:
Docs in https://github.com/gotenberg/gotenberg.dev/pull/39

PR #1457, opened 2026-01-23
fix(api): propagate request context to stop processing on client disconnect

### LINKED ISSUE #1455: Propagate HTTP request context cancellation to processing modules to save resources
**Is your feature request related to a problem?**
Yes. Currently, if a client (HTTP client) disconnects abruptly or times out while Gotenberg is processing a request (e.g., converting a large PDF or waiting for a page to load), the server continues to process the request until completion or until the hard server-side timeout is reached. This wastes CPU and RAM on tasks where the result will inevitably be discarded because the client is no longer listening. It also delays the processing of subsequent requests unnecessarily.

This behavior occurs because the context used for processing is explicitly detached from the HTTP request's lifecycle. In `pkg/modules/api/context.go`, the `newContext` function initializes the processing context using `context.Background()` instead of the request's context:

```go
// pkg/modules/api/context.go

func newContext(echoCtx echo.Context, ...) (*Context, context.CancelFunc, error) {
    // ...
    // The context is rooted in Background, effectively detaching it from the client connection.
    processCtx, processCancel := context.WithTimeout(context.Background(), timeout)
    
    ctx := &Context{
        // ...
        Context:     processCtx,
    }
    // ...
}

```

Although middleware in `pkg/modules/api/middlewares.go` defers a `cancel()` call, this only executes *after* the handler returns. Since the handler is blocking on the detached `processCtx`, it remains unaware that the client has disconnected.

**Describe the solution you'd like**
I would like the processing context to inherit from the HTTP request context so that client-side cancellations propagate immediately to the underlying modules (Chromium, LibreOffice, etc.).

If the client closes the connection:

1. `echoCtx.Request().Context().Done()` should trigger.
2. This should cause the `Context` passed to `chromium.Pdf` (and others) to be canceled.
3. The modules, which already listen to `ctx.Done()`, will abort their subprocesses early.

**Describe alternatives you've considered**
The current alternative is relying on `--api-timeout`. However, this is a "hard" limit and does not help release resources dynamically when a client gives up early (e.g., a user closing a browser tab waiting for a report).

**Additional context**
Relevant code sections:

* `pkg/modules/api/context.go`: The `newContext` function creates the detached `context.Background()`.
* `pkg/modules/api/middlewares.go`: The `contextMiddleware` passes this detached context to downstream handlers.

romanek-adam-b2c2:
Fixes #1455.

This PR links the internal processing context to the HTTP request context. Currently, `newContext` uses `context.Background()`, which creates a detached context that ignores client disconnections. By switching to `echoCtx.Request().Context()`, the `Context` passed to PDF engines and other modules will be canceled immediately if the client closes the connection (e.g., due to a timeout).

This ensures that resource-intensive operations (Chromium, LibreOffice, etc.) abort early when the client is no longer listening, freeing up the worker queue for other requests.

### DIFF
--- a/pkg/modules/api/context_test.go
+++ b/pkg/modules/api/context_test.go
@@ -27,7 +27,7 @@ func TestNewContext_Cancellation(t *testing.T) {
 		t.Fatalf("failed to close multipart writer: %v", err)
 	}
 
-    // Create a request with a cancellable context.
+	// Create a request with a cancellable context.
 	reqCtx, cancelReq := context.WithCancel(context.Background())
 	req := httptest.NewRequest(http.MethodPost, "/", body).WithContext(reqCtx)
 	req.Header.Set("Content-Type", writer.FormDataContentType())
@@ -48,7 +48,7 @@ func TestNewContext_Cancellation(t *testing.T) {
 	}
 	defer cancel()
 
-    // Verify initial state: context SHOULD NOT be done yet.
+	// Verify initial state: context SHOULD NOT be done yet.
 	select {
 	case <-ctx.Done():
 		t.Fatal("context should not be done immediately")
--- a/pkg/modules/webhook/middleware.go
+++ b/pkg/modules/webhook/middleware.go
@@ -272,6 +272,29 @@ func webhookMiddleware(w *Webhook) api.Middleware {
 						})
 						return c.NoContent(http.StatusNoContent)
 					}
+
+					if deadline, ok := ctx.Deadline(); ok {
+						// Create a new context derived from Background (detached from Request)
+						// but with the same deadline as the original context.
+						detachedCtx, detachedCancel := context.WithDeadline(context.Background(), deadline)
+
+						// Replace the embedded context in the api.Context struct.
+						// The modules downstream will now use this detached context.
+						ctx.Context = detachedCtx
+
+						// We must wrap the cancel function.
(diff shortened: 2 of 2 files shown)


gulien:
@romanek-adam-b2c2 thanks for the PR!

Ok, so this does not work with the webhook feature. 

As the request finishes early when using a webhook, it cancels the context, and therefore the asynchronous operations are also cancelled.

I wonder what can be done here, as the webhook feature lives in its own package, providing a middleware.
Maybe override somehow the `*api.Context`?

https://github.com/gotenberg/gotenberg/blob/3220ca41408bde2fae0f38aea9dbec4f17cb71f3/pkg/modules/webhook/middleware.go#L98

romanek-adam-b2c2:
Hey @gulien , I knew it wouldn't be so easy ;)

Pushed another commit to hopefully fix the issue. I also run integration tests for the webhook feature locally and they passed.

To be completely honest - I have no Go experience and all those changes are suggestions from Gemini. I think it's on the right track, but please check the code carefully 🙏

gulien:
Thanks for the help @romanek-adam-b2c2 🙏

romanek-adam-b2c2:
It was my pleasure!

romanek-adam-b2c2:
@gulien how soon are you expecting to cut a release with this change?

gulien:
This week!

PR #1467, opened 2026-02-08
Add concurrency support for Chromium

TomBrouws:
This MR updates Gotenberg to support concurrency on the Chromium module.

For now I've defaulted the flag to 1 concurrent request to keep current behaviour.

@gulien happy to make changes where necessary, and I'm also open to submit a PR to the docs repo if/when this gets merged. 🙏

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,7 @@ API_DISABLE_HEALTH_CHECK_LOGGING=false
 API_ENABLE_DEBUG_ROUTE=false
 CHROMIUM_RESTART_AFTER=100
 CHROMIUM_MAX_QUEUE_SIZE=0
+CHROMIUM_MAX_CONCURRENCY=6
 CHROMIUM_AUTO_START=false
 CHROMIUM_START_TIMEOUT=20s
 CHROMIUM_ALLOW_INSECURE_LOCALHOST=false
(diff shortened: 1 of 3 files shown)


TomBrouws:
By the way, I think we should either increase the `chromium-restart-after` default value, or advise through docs to increase it when using the Chromium concurrency feature. With the current default, the restarts happen very often and significantly slow down the entire process. Curious to hear your thoughts.

pkg/gotenberg/supervisor.go
```diff
@@ -197,33 +201,43 @@ func (s *processSupervisor) Run(ctx context.Context, logger *zap.Logger, task fu
...
 			select {
-			case s.mutexChan <- struct{}{}:
+			case s.semaphore <- struct{}{}:
 				logger.Debug("process lock acquired")
+
+				// If a restart drain is in progress, release the slot
+				// immediately so the drain can acquire it instead.
+				if s.isRestarting.Load() {
+					<-s.semaphore
+					return ErrProcessAlreadyRestarting
```
gulien: What happens is such a scenario? The request fails?

gulien:
> By the way, I think we should either increase the chromium-restart-after default value, or advise through docs to increase it when using the Chromium concurrency feature. With the current default, the restarts happen very often and significantly slow down the entire process. Curious to hear your thoughts.

We could increase the chromium-restart-after default value and set the concurrency to the max value by default, as it has more benefits for the end user, and it's not really a breaking change.

pkg/gotenberg/supervisor.go
TomBrouws: No, in lines 276-280 we catch this specific error and retry. Once all slots have been acquired by the restart mechanism and the restart has completed, this request will again be able to get a slot in line 204 and be handled as intended.

I do see I missed a small thing, I'm again increasing the queue size there which I shouldn't, will fix that.

TomBrouws:
> > By the way, I think we should either increase the chromium-restart-after default value, or advise through docs to increase it when using the Chromium concurrency feature. With the current default, the restarts happen very often and significantly slow down the entire process. Curious to hear your thoughts.
> 
> We could increase the chromium-restart-after default value and set the concurrency to the max value by default, as it has more benefits for the end user, and it's not really a breaking change.

Sure, happy to do that! Let me make the change.

TomBrouws:
@gulien I set `chromium-restart-after` to 100, and `chromium-max-concurrency` to 6

TomBrouws:
@gulien I fixed the formatting for the integration test feature file. Sorry, didn't see and run that command locally when first submitting the MR. I think you'll have to retrigger the workflow.

gulien:
Thanks @TomBrouws

PR #1490, opened 2026-03-12
fix(security): avoid shell interpolation

toindev:
Avoid possible shell interpolation of inputs by using environment variables.


gulien:
Thanks @toindev!

PR #1529, opened 2026-04-21
chore(npm): Restrict npm installations

pieterocp:
# What

Restrict npm installs + package bumps for a week & prevents scripts running in Ci.

# Why

See hacks like the [axios malware](https://www.stepsecurity.io/blog/axios-compromised-on-npm-malicious-versions-drop-remote-access-trojan), where if this project did have axios, local setup of npm didn't have it or dependabot was running, it might've lead to credentials or github account compromised which would be bad.

(afaik this isn't a backwards incompatible change since this is just for dependabot / gha / development machine changes.)


gulien:
Thanks 🤟

PR #1533, opened 2026-04-22
feat(chromium): add --chromium-disable-private-network-access flag

lewis-green:
Chrome's Private Network Access feature blocks headless browsers from fetching sub-resources hosted on private IPs when the initiating page is served from a public origin. This causes silent fetch failures for Gotenberg deployments that render pages which load data from internal REST APIs (e.g. Kubernetes ClusterIP services or Docker-internal hosts).

Add a --chromium-disable-private-network-access flag (default false) that disables the PrivateNetworkAccessSendPreflights and BlockInsecurePrivateNetworkRequests Chrome features when set, allowing those fetches to proceed. The flag is opt-in so existing deployments are unaffected.


gulien:
Not relevant anymore :) found another solution.

gulien:
And thanks @lewis-green!

PR #1543, opened 2026-05-04
feat(chromium): allow to specify device scale ratio for screenshots

hovcharenko:
### Summary
Adds a new `deviceScaleFactor` form field to Chromium with backwards-compatible defaults.

### Why
I had a need to generate hi-resolution images for sharing on social media, was not able to achieve this with currently present features.


gulien:
Thanks @hovcharenko 🙏

Issue #1136, opened 2025-02-28
Docker images on ghcr.io


ondrejsika:
What about to publish docker images also to ghcr.io?

gulien:
Hello @ondrejsika,

I don't want to increase the number of registries, but do you have a use case for ghrc.io instead of Docker hub?

ondrejsika:
Docker Hub has very low limits for anonymous image pulls.

gulien:
This would require some updates from the ci perspective but why not!

zombiehoffa:
This  is killing me right now so would be extremely appreciated if you could add ghcr.io. It' s not a bad idea to maintain two repos for reliability in case docker hub goes down or github starts acting more like its parent company....

caesay:
I echo the previous comments, for people with homelabs with many containers, dockerhub is basically unusable without paying for a pro subscription.

Since your builds are already in github actions, might be quite simple to add support for this.

Add this at the top of your workflow to give it permission to publish packages
```yaml
permissions:
  contents: read
  packages: write
```

Then add a login step somewhere, probably in your (already centralised) build-test-push action. 
```yml
      - name: Log in to GHCR
        uses: docker/login-action@v3
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}  # provided automatically by github, no need to create
```

I think you'd then only need to add some extra tags to the bit that calls your `push.sh` script. e.g. 

```yml
    - name: Push
      if: inputs.docker_hub_username != ''
      shell: bash
      run: |
        .github/actions/build-test-push/push.sh \
          --tags "ghcr.io/${{ steps.build.outputs.tags }},${{ steps.build.outputs.tags }},${{ steps.build.outputs.tags_cloud_run }},${{ steps.build.outputs.tags_aws_lambda }}" \
          --dry-run "${{ inputs.dry_run }}"
```

gulien:
Hello @caesay 👋 

Gotenberg is part of the Docker Sponsored Open Source Program, so the usual limitations do not apply (e.g. rate limit etc.)

christaikobo:
> Hello [@caesay](https://github.com/caesay) 👋
> 
> Gotenberg is part of the Docker Sponsored Open Source Program, so the usual limitations do not apply (e.g. rate limit etc.)

[...]

PR #1569, opened 2026-06-03
ci: Add GitHub Container Registry (ghcr.io) publishing

terafin:
## Summary
This PR adds GitHub Container Registry (ghcr.io) as an additional publishing target alongside Docker Hub.

## Motivation
Docker Hub's rate limiting (100 pulls/6hrs anonymous, 200 free) increasingly impacts CI/CD and self-hosted infrastructure. ghcr.io provides no rate limits for public images, a unified code+containers ecosystem, needs no extra secrets (uses the existing GITHUB_TOKEN), and reuses the same build — just an additional registry target.

## Changes
This change deliberately reuses the repository's *existing* alternate-registry mechanism rather than adding any new build or tag logic:

- `.github/workflows/continuous-delivery.yml`:
  - Added `permissions: contents: read` + `packages: write` to the `merge_clean_release_tags` job so it can push to ghcr.io.
  - Passed `alternate_registry: ghcr.io/${{ github.repository_owner }}` to each of the existing `Merge` steps (full, Chromium, LibreOffice, AWS Lambda, AWS Lambda Chromium, AWS Lambda LibreOffice). This produces `ghcr.io/${{ github.repository }}` (i.e. `ghcr.io/gotenberg/gotenberg`) tags via the already-present `alternate_registry` path in `.github/actions/merge/merge.sh`.
- `.github/actions/merge/action.yml`:
  - Added a `docker/login-action@v3` step for `ghcr.io` (username `${{ github.actor }}`, password `${{ inputs.github_token }}`) immediately after the existing Docker Hub login. It is guarded by `if: inputs.alternate_registry != ''`, so it only runs when an alternate registry is requested.

No build args, platforms/multi-arch, build context, cache, tag strategy, or existing Docker Hub tags were changed. The ghcr.io manifests are created by re-tagging the already-merged multi-arch Docker Hub manifest via `docker buildx imagetools create`, so they carry the identical set of platforms and the same tag suffixes.

## Backward Compatibility
Fully backward compatible — Docker Hub publishing is unchanged; this only adds an additional registry target.

## Note on Previous Discussion
I saw this was discussed previously and declined due to Docker Sponsored OSS status. I wanted to revisit because: (1) many users run private CI/CD that doesn't benefit from Docker Sponsored OSS exemptions; (2) self-hosted environments (Kubernetes, Docker Swarm) still hit rate limits; (3) ghcr.io adds redundancy if Docker Hub has availability issues. Completely understand if the answer is still no — happy to close if not wanted. (Refs #1136.)

## Testing
- [x] Workflow YAML validated
- [ ] Builds in maintainer CI on merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)
---

### 🔧 One-time maintainer step: make the GHCR package public

Heads-up for maintainers: the first time this workflow publishes to `ghcr.io/gotenberg/gotenberg`, GitHub creates the package as **private** by default. To let users `docker pull` it without authentication, a maintainer needs to set its visibility to **Public** once:

> Repo **Packages** → the new `gotenberg` package → **Package settings** → **Danger Zone** → **Change visibility** → **Public**

It's a one-time action — subsequent pushes inherit the setting. (Flagged by an automated reviewer; surfacing it here so the rollout is smooth.)

### DIFF
--- a/.github/actions/merge/action.yml
+++ b/.github/actions/merge/action.yml
@@ -39,6 +39,14 @@ runs:
         username: ${{ inputs.docker_hub_username }}
         password: ${{ inputs.docker_hub_password }}
 
+    - name: Log in to GitHub Container Registry
+      if: inputs.alternate_registry != ''
+      uses: docker/login-action@v3
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ inputs.github_token }}
+
     - name: Merge
       shell: bash
       env:
--- a/.github/workflows/continuous-delivery.yml
+++ b/.github/workflows/continuous-delivery.yml
@@ -152,6 +152,9 @@ jobs:
       - release_arm_v7
     name: Merge and clean release tags
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
     steps:
       - name: Checkout source code
         uses: actions/checkout@v6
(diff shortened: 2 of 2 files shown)


gulien:
Thanks @terafin! I really don't want to have many registries, and prefer to keep it in Docker Hub.

PR #1573, opened 2026-06-05
fix(Dockerfile): add ca-certificates, missing in chromium-only image

osvein:
The gotenberg and gotenberg-libreoffice images pull in ca-certificates as a dependency of python3-distutils-extra, but it's missing from the gotenberg-chromium image. The issue only affects webhooks, as Chromium itself uses bundled certificates.


gulien:
Thanks!

PR #1586, opened 2026-06-24
feat(poppler): add PDF to image conversion route

falzarain:
Add a new poppler module wrapping pdftoppm to rasterize PDFs into images, exposed at POST /forms/pdfengines/convert/image. Each page becomes a PNG, JPEG, or TIFF; a single page returns the image directly while multiple pages return a ZIP archive.

The module follows the qpdf pattern: it reads PDFTOPPM_BIN_PATH, traces the binary call with an OTEL client span, and self-registers via the standard imports. poppler-utils is added to the Docker image, with its version captured at build time.

### DIFF
--- a/.bruno/PDF Engines/Convert/Convert To Image.bru
+++ b/.bruno/PDF Engines/Convert/Convert To Image.bru
@@ -1,0 +1,30 @@
+meta {
+  name: Convert To Image
+  type: http
+  seq: 2
+}
+
+post {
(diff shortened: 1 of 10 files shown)


falzarain:
Hi @gulien , any chance to accept the merge? We are already using it as a fork but i think is an useful feature for everyone

effyteva:
Looking forward for this feature

PR #1613, opened 2026-07-29
fix(libreoffice): return 500 for unexplained UNO errors

### LINKED ISSUE #1588: convert to PDF: supervisor run task: uno exception returns 400 status code instead of 500
Hi, we've been seeing the following error intermittently when converting large documents:

convert to PDF: supervisor run task: uno exception

We're aware this is likely related to LibreOffice struggling with large files under memory pressure (also mentioned in [#1399](https://github.com/gotenberg/gotenberg/issues/1399)), and we understand the recommended fix is to increase memory/CPU limits.

That said, we wanted to flag what feels like a potential issue with the HTTP status code returned in this case. Gotenberg returns 400 Bad Request, but since the request itself is valid, we'd expect a 500 here. The [endpoint docs](https://gotenberg.dev/docs/convert-with-libreoffice/convert-to-pdf#basics) also describe 400 as being for invalid client requests.

The reason this matters to us is that we rely on status codes to decide whether to retry a failed conversion. A 400 tells us "your request was bad, don't bother retrying," which isn't really the right signal for a server-side crash. Happy to hear if there's a reason for this behavior we might be missing!

log entry:
```javascript
{
  "message": "convert to PDF: supervisor run task: uno exception",
  "method": "POST",
  "path": "/forms/libreoffice/convert",
  "status": 400,
  "latency_human": "4.630966661s"
}
```

Nameless-Monster-Nerd:
## Summary

- return unexplained LibreOffice UNO failures as internal server errors
- preserve HTTP 400 for UNO failures associated with client-supplied page ranges or passwords
- add table-driven coverage for all three paths

## Testing

- `make test-unit`
- `go test ./pkg/modules/libreoffice/...`
- `gofmt` on changed Go files
- `git diff --check`

Fixes #1588.

### DIFF
--- a/pkg/modules/libreoffice/routes.go
+++ b/pkg/modules/libreoffice/routes.go
@@ -15,6 +15,22 @@ import (
 	"github.com/gotenberg/gotenberg/v8/pkg/modules/pdfengines"
 )
 
+func wrapUnoException(err error, options libreofficeapi.Options) error {
+	wrapped := fmt.Errorf("convert to PDF: %w", err)
+
+	if options.PageRanges == "" && options.Password == "" {
+		return wrapped
+	}
+
+	return api.WrapError(
+		wrapped,
+		api.NewSentinelHttpError(
+			http.StatusBadRequest,
+			fmt.Sprintf("LibreOffice failed to process a document: possible causes include malformed page ranges '%s' (nativePageRanges), or, if a password has been provided, it may not be required. In any case, the exact cause is uncertain.", options.PageRanges),
+		),
+	)
+}
+
 // convertRoute returns an [api.Route] which can convert LibreOffice documents
 // to PDF.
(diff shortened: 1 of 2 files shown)


pkg/modules/libreoffice/routes.go
```diff
@@ -15,6 +15,22 @@ import (
 	"github.com/gotenberg/gotenberg/v8/pkg/modules/pdfengines"
 )
 
+func wrapUnoException(err error, options libreofficeapi.Options) error {
```
gulien, line 18: Could you inline it in the `convertRoute` instead? 😄

gulien (changes requested):

gulien:
Done in 31fa392db270defa572dc6d629a56362827e2798

PR #1617, opened 2026-07-31
Await thenable js expression

nhh:
I want to write my wait for expressions like this and chrome seems to support this. I could not test it agains a go compiler due to my work env but it "should compile™"

```js
(async () => {
  document.querySelector('#accept').click();
  await new Promise(r => setTimeout(r, 2000)); // wait 2s
  return true;
})()
```

With this you can write more sophisticated wait for expressions in a "more clean" manner. What do you think?


gulien:
Thanks!

Issue #1634, opened 2026-08-18
Image 8.36.0 contains fixable CRITICAL/HIGH CVEs, maybe add automated scanning?


delta-whiplash:
First of all, thanks a lot for Gotenberg, I use it alongside Paperless-ngx for document conversion and it's been working great in production. Really appreciate the work you put into it! 🙏

I noticed the `gotenberg/gotenberg:8.36.0` image fails our Trivy scan with fixable CRITICAL/HIGH CVEs:

```
docker.io/gotenberg/gotenberg:8.36.0@sha256:87c16b9f364279d321bc9772d31fa58aa6abe036423c270698bd636c3a8e9466
```

Could you have a look and rebuild with patched base packages when you get a chance, please?

On a longer-term note, since you're already using GitHub Actions for releases, adding some automated tooling could prevent this class of issues from reaching published images:

- **Trivy** (or Grype) as a CI step on each image build: fail the build on fixable CRITICAL/HIGH so CVEs get caught before a tag ships
- **Dependabot or Renovate** on the repo: keeps Go module and Dockerfile base image dependencies automatically up to date via PRs
- Bonus: Renovate can even watch the base image and open a PR whenever a patched digest is published

Thanks again for the project! 🚀

PR #1635, opened 2026-08-19
chore: Use renovate to bump go and dockerfile

MarkIannucci:
Dependabot doesn't do well bumping versions specified in dockerfile [ARG values](https://github.com/dependabot/dependabot-core/issues/2057).

This PR introduces configuration for [renovate](https://github.com/renovatebot/renovate) which will simultaneously bump the go version in go.mod, the dockerfile ARG value, and the golangci-lint version as [demoed in my fork repo](https://github.com/MarkIannucci/gotenberg/pull/4/changes).

In order for it to work, a gotenberg org owner will need to install the [Renovate GitHub app](https://github.com/apps/renovate) and go through the onboarding after merging this configuration.  Alternatively, you can run renovate using the [Renovate GH Action](https://github.com/renovatebot/github-action) in a GH workflow.  Doing that requires obtaining a `GH_TOKEN` and passing it to the action.  If the maintainers would prefer to go that route, I'm happy to write the action.

Additionally, it is possible to use renovate for all of the deps and delete the dependabot configuration.  I'm happy to do that if the maintainers would prefer.

Partially completes #1634.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,10 +4,6 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-  - package-ecosystem: "docker"
-    directory: "/build"
-    schedule:
-      interval: "weekly"
   - package-ecosystem: "gomod"
     directory: "/"
     schedule:
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,47 @@
+{
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": ["config:recommended"],
+  "enabledManagers": ["dockerfile", "gomod", "custom.regex"],
+  "semanticCommits": "enabled",
+  "schedule": ["before 6am on monday"],
+  "packageRules": [
+    {
+      "description": "Dependabot owns Go module dependencies. Renovate only handles the toolchain.",
+      "matchManagers": ["gomod"],
+      "enabled": false
+    },
+    {
+      "description": "Bump the go directive in go.mod. Renovate skips it by default.",
+      "matchManagers": ["gomod"],
+      "matchDepTypes": ["golang"],
+      "matchDepNames": ["go"],
+      "enabled": true,
+      "rangeStrategy": "bump",
+      "groupName": "Go toolchain"
+    },
+    {
+      "description": "Keep ARG GOLANG_VERSION in step with the go directive, in a single PR.",
+      "matchManagers": ["dockerfile"],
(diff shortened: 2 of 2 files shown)


MarkIannucci:
Converting to draft.  This didn't take care of the latest version of go correctly in markiannucci/gotenberg#4.  Will take a deeper look later tonight and hopefully improve.

MarkIannucci:
Rebased and added functionality to get it also bump golangci-lint so that way CI would keep up with the version of go we use.

timvandam:
hi @gulien, any chance this PR could be merged into gotenberg? would help auto fix CVEs in gotenberg images. for context, images with CVEs get blocked by many deploy policies

gulien:
Hello @timvandam,

Thanks for the PR! I'm not yet familiar enough with the renovate bot, so I need time to educate myself.

gulien:
Hello @MarkIannucci,

Thanks for the work on this, and for coming back to it twice. Your diagnosis is correct, and I checked it against the history before replying: Dependabot has opened 127 PRs on this repo and not one came from the `docker` ecosystem. Every Dockerfile version bump here is a manual commit of mine. The blind spot is real.

I am going to close this anyway, so here is the reasoning rather than just a no.

The cost is a second bot running alongside Dependabot, and the gap it would close is narrower than it looks. Dependabot already covers Go, npm and Actions. What is left is a handful of pins in the Dockerfile, and I already go through those by hand when preparing a release.

Scope is the part that decided it for me. The pins that actually drift are `PDFCPU_VERSION`, `PDFTK_VERSION`, the lambda adapter, `TMP_CHROMIUM_VERSION_PPC64EL`, and unoconverter, which is not even an ARG, just a version inside a curl URL. This PR covers the Go toolchain and golangci-lint, which are the two I am least likely to let slip.

One expectation to set, since #1634 is linked and @timvandam raised CVEs: merging this would not have moved that needle. Those findings come from Debian packages, and every release rebuilds the image with whatever is current. I also do not ship releases just to patch dependencies, because a given tag has to behave the same for everyone. Beyond that, a lot of what scanners report has no fix available.

Worth knowing as well: v9 will move to a custom PDF engine, no ETA. That removes pdftk and its bundled JRE, qpdf, exiftool and pdfcpu from the image, so a good part of the surface being scanned today disappears on its own. Another reason I do not want to build tooling around it right now.

None of this is a closed door on dependency work. PRs for dependency updates and package patches are welcome any time. Thanks again 👍

PR #1641, opened 2026-08-28
fix(chromium): fail fast with 503 when Chromium crashes

### LINKED ISSUE #1640: Fail fast on Chromium crash
I ran into the same issue in #545 and in our case the reason is in the AVIF render of Chromium in the arm64 build.

However, even when Chromium crashes, the offending request should not timeout, but directly fail with 503.

Haseeb-1698:
Closes #1640

When the renderer crashes mid-conversion, the in-flight CDP command never receives a response, so the request hangs until the API deadline and surfaces as a generic timeout. This listens for `Inspector.targetCrashed` and maps it to an explicit 503 through the existing sentinel-error path.

A dead browser context after a failed run is treated the same way. That check is gated on `runErr != nil`, so a successful conversion is never discarded by a browser death landing just after it.

## Verification

Same request (`url=chrome://crash`, `CHROMIUM_ALLOW_LIST='.+'`), before and after:

| Build | Result |
|---|---|
| `main` | **30.01s** — 503, generic request-timeout body |
| This PR | **0.56s** — 503, crash body |

The service recovers normally: the next conversion returns 200 with a valid PDF in 0.12s.

`chrome://crash` is a deterministic renderer-crash simulation on x86_64. I could not reproduce the reporter's arm64/AVIF trigger, but `Inspector.targetCrashed` is trigger-agnostic, so the detection path is the same.

## Tests

- `routes_test.go` — 503 mapping for the new sentinel error
- `errortype_test.go` — added the taxonomy case
- Integration scenario driving a real renderer crash and asserting the 503 and the crash body

`make fmt`, `golangci-lint` v2.12.2 (0 issues), `go test -race ./pkg/modules/chromium/...`, and the integration scenario all pass locally. No API routes were added or changed — only error mapping inside the existing handler — so the Bruno collection is unchanged.

## Notes

This targets `main` (v8). I noticed #1487 reworks this module for Gotenberg 9 — happy to port the same change there as well if that would be useful.

It also touches `chromium.go` and `routes.go`, which #1639 modifies; glad to rebase if that lands first.

### DIFF
--- a/pkg/modules/chromium/browser.go
+++ b/pkg/modules/chromium/browser.go
@@ -523,7 +523,44 @@ func (b *chromiumBrowser) do(ctx context.Context, logger *slog.Logger, url strin
 		cancelOnMainPageError:   taskCancel,
 	})
 
+	var (
+		crashed   error
+		crashedMu sync.RWMutex
+	)
+
+	// See https://github.com/gotenberg/gotenberg/issues/1640.
+	listenForEventTargetCrashed(taskCtx, logger, eventTargetCrashedOptions{
+		crashed:   &crashed,
+		crashedMu: &crashedMu,
+		cancel:    taskCancel,
+	})
+
 	runErr := chromedp.Run(taskCtx, tasks...)
+
+	// A crashed renderer is the root cause of every other failure this
+	// conversion may have recorded, so check it first.
+	// See https://github.com/gotenberg/gotenberg/issues/1640.
+	crashedMu.RLock()
+	defer crashedMu.RUnlock()
+
+	if crashed != nil {
+		return fmt.Errorf("handle tasks: %w", crashed)
+	}
+
+	// The browser context is only ever canceled when the browser process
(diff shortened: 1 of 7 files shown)


Haseeb-1698:
CI caught a flaky assertion of mine on `linux/ppc64le`. I had asserted the response completes within 10s; under QEMU emulation Chromium's cold start alone pushes the request to ~13s, so it failed there. Every functional assertion passed on that architecture — 503, `Content-Type`, and the crash body — and all other jobs were green, including lint and unit tests.

I've removed the wall-clock assertion and the `scenario.go` step that backed it. The body assertion already proves the behaviour: `Chromium crashed while processing the request` can only come from the new path, since before this change the request fell through to the API timeout and returned a different message. That also takes `scenario.go` out of the diff entirely — now 7 files, +154/−0.

The before/after timing evidence stays in the description above.

gulien:
Thanks!

PR #1645, opened 2026-09-01
fix(Dockerfile): re-add libreoffice-math so DOCX math formulas render

### LINKED ISSUE #1644: DOCX→PDF silently drops all math formulas since 8.30.0 — `libreoffice-math` was removed in the image slimming
## Summary

Since v8.30.0, `POST /forms/libreoffice/convert` silently omits every OMML math formula (`m:oMath`, the standard Word equation format) from converted DOCX files. The request returns `200` with a structurally valid PDF — the formulas are just gone, including any formula-only paragraphs. Surrounding text is unaffected, so the loss is easy to miss until an end user compares the PDF against the source document.

Root cause: v8.30.0's Dockerfile rework (9e5b05ad, f8809b39) replaced the Debian `libreoffice` metapackage with an explicit component list — `libreoffice-writer libreoffice-calc libreoffice-impress libreoffice-draw python3-uno` — in **both** the full and `-libreoffice` final stages. That list omits `libreoffice-math`. Writer imports OMML as embedded Math objects, and without the Math component installed it renders them as nothing.

This is not a LibreOffice bug: the same LibreOffice version (26.2.2.2, official TDF build) on plain Debian converts the same file correctly with `soffice --convert-to pdf`.

## Affected versions

- v8.29.1 and earlier: OK (installs the `libreoffice` metapackage, which includes Math)
- v8.30.0 through v8.36.0: broken, all variants (full and `-libreoffice` behave identically)
- `main` today still has the same package list, so the next release will be broken too

## Reproduce

Minimal DOCX (one fraction formula between two text paragraphs), built from scratch so no proprietary document is needed:

```bash
python3 - <<'EOF'
import zipfile

document = '''<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<w:document xmlns:w="http://schemas.openxmlformats.org/wordprocessingml/2006/main"
            xmlns:m="http://schemas.openxmlformats.org/officeDocument/2006/math">
  <w:body>
    <w:p><w:r><w:t>Text before the formula.</w:t></w:r></w:p>
    <w:p>
      <m:oMathPara><m:oMath>
        <m:r><w:rPr><w:rFonts w:ascii="Cambria Math" w:hAnsi="Cambria Math"/></w:rPr><m:t>E=</m:t></m:r>
        <m:f>
          <m:num><m:r><w:rPr><w:rFonts w:ascii="Cambria Math" w:hAnsi="Cambria Math"/></w:rPr><m:t>a+b</m:t></m:r></m:num>
          <m:den><m:r><w:rPr><w:rFonts w:ascii="Cambria Math" w:hAnsi="Cambria Math"/></w:rPr><m:t>c</m:t></m:r></m:den>
        </m:f>
      </m:oMath></m:oMathPara>
    </w:p>
    <w:p><w:r><w:t>Text after the formula.</w:t></w:r></w:p>
  </w:body>
</w:document>'''

content_types = '''<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">
  <Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>
  <Default Extension="xml" ContentType="application/xml"/>
  <Override PartName="/word/document.xml" ContentType="application/vnd.openxmlformats-officedocument.wordprocessingml.document.main+xml"/>
</Types>'''

rels = '''<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
  <Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" Target="word/document.xml"/>
</Relationships>'''

with zipfile.ZipFile('formula.docx', 'w', zipfile.ZIP_DEFLATED) as z:
    z.writestr('[Content_Types].xml', content_types)
    z.writestr('_rels/.rels', rels)
    z.writestr('word/document.xml', document)
EOF

docker run -d --rm --name gb -p 3000:3000 gotenberg/gotenberg:8.36.0
sleep 5
curl -o out.pdf -F "files[]=@formula.docx" http://localhost:3000/forms/libreoffice/convert
docker rm -f gb
```

`out.pdf` contains "Text before the formula." and "Text after the formula." — the `E = (a+b)/c` line is missing entirely. The same request against `gotenberg/gotenberg:8.29.1` renders the formula.

## Evidence it's the missing package

```console
$ docker run --rm --entrypoint bash gotenberg/gotenberg:8.29.1 -c "dpkg -l | grep -c libreoffice-math"
1
$ docker run --rm --entrypoint bash gotenberg/gotenberg:8.34.0-libreoffice -c "dpkg -l | grep -c libreoffice-math"
0
```

Rebuilding the current image with the one package restored fixes it:

```dockerfile
FROM gotenberg/gotenberg:8.34.0-libreoffice
USER root
RUN echo 'deb http://deb.debian.org/debian trixie-backports main' > /etc/apt/sources.list.d/backports.list \
 && apt-get update -qq \
 && apt-get install -y -qq --no-install-recommends -t trixie-backports libreoffice-math \
 && rm -rf /var/lib/apt/lists/* /etc/apt/sources.list.d/backports.list
USER gotenberg
```

With that image the same request renders the formula correctly.

## Suggested fix

Add `libreoffice-math` to the package list in both final stages of `build/Dockerfile`. It's a small package (its hard dependencies are already in the image — installing it in the container adds only a few MB), and it restores behavior that every release before 8.30.0 had. The docs currently state the `-libreoffice` variant only gives up the Chromium routes, so users have no way to anticipate this loss — documents with equations are common in technical, academic, and regulated-industry use.

Possibly related: #1389 reports equation rendering discrepancies (spacing) — different symptom, but it may be worth re-checking once Math is back in the image.

joh-klein:
Fixes #1644.

Re-adds `libreoffice-math` to the LibreOffice package list in both final stages (full and `-libreoffice`). The v8.30.0 slimming switched from the `libreoffice` metapackage to an explicit component list, and Math was left out — since then every OMML equation (`m:oMath`, the standard Word format) is silently dropped from DOCX→PDF conversions: the response is still `200` with a valid PDF, the formulas are just missing.

Size impact is small: Math's hard dependencies are already in the image via the other components — the resulting `-libreoffice` image is ~1.05GB, only a few MB over the current one.

**Verification** (repro script and screenshots in #1644):

- Built the `gotenberg-libreoffice` target from this branch (`make build TARGET=gotenberg-libreoffice`, linux/arm64): `dpkg -l` shows `libreoffice-math` installed, and the repro DOCX from #1644 renders its `E = (a+b)/c` formula correctly via `POST /forms/libreoffice/convert`. On the stock `8.36.0` image the formula is absent.
- Independently cross-checked by extending the published `8.34.0-libreoffice` image with only `apt-get install -t trixie-backports libreoffice-math` — same result.

Happy to add a Gherkin scenario with a formula DOCX under `test/integration/` if you'd like one — just point me at the preferred assertion style for content-level checks.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -259,7 +259,7 @@ RUN echo "deb http://deb.debian.org/debian trixie-backports main" >> /etc/apt/so
         hyphen-no hyphen-or hyphen-pa hyphen-pl hyphen-pt-br hyphen-pt-pt hyphen-ro hyphen-ru hyphen-sk hyphen-sl \
         hyphen-sr hyphen-sv hyphen-ta hyphen-te hyphen-th hyphen-uk hyphen-zu \
     && DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends -t trixie-backports \
-        libreoffice-writer libreoffice-calc libreoffice-impress libreoffice-draw python3-uno \
+        libreoffice-writer libreoffice-calc libreoffice-impress libreoffice-draw libreoffice-math python3-uno \
     # unoconverter will look for the Python binary, which has to be at version 3.
     && ln -s /usr/bin/python3 /usr/bin/python \
     # Cleanup.
@@ -386,7 +386,7 @@ RUN echo "deb http://deb.debian.org/debian trixie-backports main" >> /etc/apt/so
         hyphen-no hyphen-or hyphen-pa hyphen-pl hyphen-pt-br hyphen-pt-pt hyphen-ro hyphen-ru hyphen-sk hyphen-sl \
         hyphen-sr hyphen-sv hyphen-ta hyphen-te hyphen-th hyphen-uk hyphen-zu \
     && DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends -t trixie-backports \
-        libreoffice-writer libreoffice-calc libreoffice-impress libreoffice-draw python3-uno \
+        libreoffice-writer libreoffice-calc libreoffice-impress libreoffice-draw libreoffice-math python3-uno \
     # unoconverter will look for the Python binary, which has to be at version 3.
     && ln -s /usr/bin/python3 /usr/bin/python \
     # Cleanup.


gulien:
Thanks!

